PR #54, opened 2017-01-05
CA-80, CA-82, CA-84

marcbeermann:
CA-80 StartProjectViewModel, CA-82 DisplayProjectViewModel, CA-84 BrowseProjectsViewModel

- The Create Project Action actually uses the new EditProjectViewModel (as does the Edit Project Action).
- Added validation when user attempts to create a project of the same name.
- Made changes to ProjectsController to calculate the participation counts (needed for 'Progress' field)
- Added 'RemainingDays' and 'StatusDescription' properties to the Project model.

### DIFF
--- a/src/CollAction/Controllers/ProjectsController.cs
+++ b/src/CollAction/Controllers/ProjectsController.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Linq.Expressions;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Rendering;
@@ -37,20 +36,28 @@ namespace CollAction.Controllers
         // GET: Project/Find
         public async Task<IActionResult> Find(FindProjectsViewModel model)
         {
-            if (model.SearchText == null) {
-                return View(new FindProjectsViewModel {
-                    Projects = await GetProjectDisplayItemsWhere(p => p.Status != ProjectStatus.Hidden)
+            if (model.SearchText == null)
+            {
+                return View(new FindProjectsViewModel
+                {
+                    OwnerId = (await _userManager.GetUserAsync(User)).Id,
+                    Projects = await DisplayProjectViewModel.GetViewModelsWhere(_context, p => p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted)
                 });
             }
 
-            model.Projects = await GetProjectDisplayItemsWhere(
-                p => p.Status != ProjectStatus.Hidden && (p.Name.Contains(model.SearchText) || p.Description.Contains(model.SearchText) || p.Goal.Contains(model.SearchText)));
+            model.OwnerId = (await _userManager.GetUserAsync(User)).Id;
+            model.Projects = await DisplayProjectViewModel.GetViewModelsWhere(_context, p => p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted &&
+                (p.Name.Contains(model.SearchText) || p.Description.Contains(model.SearchText) || p.Goal.Contains(model.SearchText)));
             return View(model);
         }
 
         public async Task<IActionResult> Index()
         {
-            return View(await GetProjectDisplayItemsWhere(p => p.Status != ProjectStatus.Hidden));
+            return View(new BrowseProjectsViewModel
+            {
+                OwnerId = (await _userManager.GetUserAsync(User)).Id,
+                Projects = await DisplayProjectViewModel.GetViewModelsWhere(_context, p => p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted)
+            });
         }
 
         // GET: Projects/Details/5
@@ -61,7 +68,7 @@ namespace CollAction.Controllers
                 return NotFound();
             }
 
-            List<DisplayProjectViewModel> items = await GetProjectDisplayItemsWhere(p => p.Id == id && p.Status != ProjectStatus.Hidden);
+            List<DisplayProjectViewModel> items = await DisplayProjectViewModel.GetViewModelsWhere(_context, p => p.Id == id && p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted);
             if (items.Count == 0)
             {
                 return NotFound();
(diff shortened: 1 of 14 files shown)


src/CollAction/Controllers/ProjectsController.cs
```diff
@@ -59,13 +61,30 @@ public ViewResult StartInfo()
...
+            if (items.Count == 0)
             {
                 return NotFound();
             }
 
-            return View(project);
+            return View(items.First());
+        }
+
+        public async Task<List<DisplayProjectViewModel>> GetProjectDisplayItemsWhere(Expression<Func<Project, bool>> WhereExpression)
```
stewartmatheson: Does this method belong in the controller? Coming from a rails background I would look to express this somewhere else but if this is an ASP.NET thing to do then that's fine. Don't let it block the pull request.

stewartmatheson:
Oh also I would have one pull request per ticket as well as one ticket per commit. Makes history easier to understand down the track. Again should not block anything just something worth noting.

src/CollAction/Models/Project.cs
```diff
@@ -67,11 +67,32 @@ public bool IsClosed
...
         public string HashTags
             => string.Join(";", Tags?.Select(tag => tag.Tag.Name) ?? Enumerable.Empty<string>());
 
+        [NotMapped]
+        public int RemainingDays
+            => Convert.ToInt32(Math.Round((End - Start).TotalDays));
+
+        [NotMapped]
+        [Display(Name = "Status")]
+        public string StatusDescription
```
logophobia: A more appropiate place would be to put it in the viewmodel. It's purely related to how to display a project, not the project data itself. It's also not localized. In fact, if you put this property here, it would also be impossible to localize in the future.

src/CollAction/Models/ProjectViewModels/DisplayProjectViewModel.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace CollAction.Models
+{
+    public class DisplayProjectViewModel
+    {
+        public Project Project;
```
logophobia: A small style issue, use:
public Project Project { get; set; }
for this, a shorthand for a getter. If you make it a property, then it's possible to change the implementation and storage of the property without changing the interface.

It doesn't matter terribly much for viewmodel classes, still, good habit to get into. https://msdn.microsoft.com/en-us/library/x9fsa0sw.aspx

src/CollAction/Models/ProjectViewModels/DisplayProjectViewModel.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        public int Participants;
```
logophobia: Same as above.

src/CollAction/Controllers/ProjectsController.cs
```diff
@@ -34,21 +35,22 @@ public ViewResult StartInfo()
         }
 
         // GET: Project/Find
-        public async Task<IActionResult> Find(FindProjectViewModel model)
+        public async Task<IActionResult> Find(FindProjectsViewModel model)
         {
             if (model.SearchText == null) {
-                return View(new FindProjectViewModel { Projects = await _context.Projects.ToListAsync() });
+                return View(new FindProjectsViewModel {
+                    Projects = await GetProjectDisplayItemsWhere(p => p.Status != ProjectStatus.Hidden)
```
logophobia: Status deleted also has to be filtered

src/CollAction/Controllers/ProjectsController.cs
```diff
@@ -34,21 +35,22 @@ public ViewResult StartInfo()
...
+                });
             }
 
-            model.Projects = await _context.Projects.Where(p => p.Name.Contains(model.SearchText) || p.Description.Contains(model.SearchText) || p.Goal.Contains(model.SearchText)).ToListAsync();
-            
+            model.Projects = await GetProjectDisplayItemsWhere(
+                p => p.Status != ProjectStatus.Hidden && (p.Name.Contains(model.SearchText) || p.Description.Contains(model.SearchText) || p.Goal.Contains(model.SearchText)));
```
logophobia: Filter status deleted

src/CollAction/Models/Project.cs
marcbeermann: That makes a lot of sense, I've done another commit with this moved to the DisplayProjectViewModel.

src/CollAction/Views/Projects/Create.cshtml
```diff
@@ -2,17 +2,17 @@
 
 @inject IViewLocalizer Localizer
 
-@model CollAction.Models.CreateProjectViewModel
```
logophobia: Right now CreateProject and EditProject use the same data. I can imagine it'd be temping to unify the viewmodels, but this feels rather premature.

I'm almost certain that in the future we'll do things slightly different. A project that's being editted for example:
* Might not be allowed to change the start date retroactively
* Might be able to administrate comments, block unwanted users (just fictional examples). Stuff that's not relevant to CreateProject

Tying them together means changing something in the edit workflow, might break create, and retrospectively.

Trying to unify stuff like this should be done at the end of the project, when we know more about how the final project looks.

src/CollAction/Views/Projects/Index.cshtml
```diff
@@ -47,33 +53,39 @@
...
             </td>
             <td>
-                @Html.DisplayFor(content => item.End)
+                @Html.DisplayFor(content => item.Progress)%
             </td>
             <td>
-                <a asp-action="Edit" asp-route-id="@item.Id">Edit</a> |
-                <a asp-action="Details" asp-route-id="@item.Id">Details</a> |
-                <a asp-action="Delete" asp-route-id="@item.Id">Delete</a>
+                <a asp-action="Edit" asp-route-id="@item.Project.Id">@Localizer["Edit"]</a> |
```
logophobia: This is still stuff left over from the auto-generated project views that asp.net can generate for you. I don't think it makes a lot of sense to show edit and delete buttons to everyone in all the project lists.

src/CollAction/Controllers/ProjectsController.cs
```diff
@@ -84,21 +103,21 @@ public ViewResult StartInfo()
         [HttpPost]
         [ValidateAntiForgeryToken]
         [Authorize]
-        public async Task<IActionResult> Create([Bind("Name,Description,Goal,CategoryId,LocationId,Target,End")] CreateProjectViewModel createProjectVM)
+        public async Task<IActionResult> Create([Bind("Name,Description,Goal,CategoryId,LocationId,Target,End")] EditProjectViewModel createProjectViewModel)
```
logophobia: EditProjectViewModel createProjectViewModel

Another issue with conflating the create and edit workflows

src/CollAction/Controllers/ProjectsController.cs
```diff
@@ -160,9 +184,15 @@ public ViewResult StartInfo()
         [HttpPost]
         [ValidateAntiForgeryToken]
         [Authorize]
-        public async Task<IActionResult> Edit(int id, [Bind("Id,Name,Description,Goal,CategoryId,LocationId,Target,End")] Project project)
+        public async Task<IActionResult> Edit(int id, [Bind("Id,OwnerId,Name,Description,Goal,CategoryId,LocationId,Target,End")] EditProjectViewModel editProjectViewModel)
```
logophobia: Here you're taking the owner id as parameter. This means the user could change the owner-id from the form, and circumvent the owner validation! Security vulnerability.

src/CollAction/Controllers/ProjectsController.cs
```diff
@@ -110,23 +129,16 @@ public ViewResult StartInfo()
                 }
                 catch (DbUpdateException ex)
                 {
-                    if (ex.InnerException is PostgresException)
```
logophobia: * Validation this way is probably most efficient, but I really would rather put validation like this either in the viewmodel or just as a separate query at the start (AnyAsync(project => project.Name.Equals..)). I realize that might lead to rare race conditions (two people submitting projects with the same same at the exact same time), but it's more important that validation (exception code checking isn't really self-explanatory) is clear rather than a small gain in efficiency in a workflow that'd only be used when creating projects. Worst case scenario, one in a million people get a generic, error creating project, and retry. But our validation is a lot more clear.
* In the future you could use exception filters for stuff where the catch(ExceptionType.. isn't sufficient

logophobia:
Overall, good stuff. Does need some tweaks.

PR #56, opened 2017-01-31
CA-119 Fixed Project.Name not editable bug.

marcbeermann:
Changed Project.Name from using an AlternateKey to using a UniqueIndex.


logophobia:
Looks good.

PR #57, opened 2017-02-01
CA175 Added 'Proposal'  and 'Comments' to Project

marcbeermann:
- Also set character limits for project name 50, proposal 300, description 1000, goal 1000 and comments 2000 as per specifications in CA-122


logophobia:
Looks good

PR #61, opened 2017-02-12
CA-155 Jobs table

minayao:


src/CollAction/Models/Job.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        [Required]
+        [MaxLength(200)]
+        public string Title { get; set; }
+
+        [Required]
+        [MaxLength(500)]
+        public string Description { get; set; }
+
+        [Required]
```
logophobia: I wouldn't make this required. Maybe there're jobs without locations?

src/CollAction/Models/Job.cs
```diff
@@ -0,0 +1,29 @@
...
+        public int LocationId { get; set; }
+        [ForeignKey("LocationId")]
+        public Location Location { get; set; }
+
+        [Required]
+        public DateTime Posting { get; set; }
```
logophobia: Naming? Maybe PostDate is clearer than Posting?

PR #62, opened 2017-02-12
Project create form

stewartmatheson:
Built project create from styles. 
Created component "starting project". 
Removed bower and it's deps and replaced boot strap with a leaner version that is just the grid system. 
Rebuilt header.
Finished footer.


logophobia:
Looks good, not an expert on react so I can't judge too much on style, but make sure to document any additional setup or build steps on the wiki: https://collaction.atlassian.net/wiki/display/CATS/Getting+CollAction+running+locally

src/CollAction/Views/Projects/Find.cshtml
```diff
@@ -9,118 +9,7 @@
     var defaultDisplayModel = FindProjectViewModel.DefaultDisplayProjectViewModel;
 }
 
-<h2>@Localizer["Find"]</h2>
```
logophobia: Hmm, this deletes most of the Find page. Is this intended. In the other commits, you replace it with different html.

src/CollAction/Views/Projects/Create.cshtml
```diff
@@ -8,89 +8,83 @@
...
         </div>
     </div>
-</form>
-
-<div>
-    <a asp-action="Index">@Localizer["Back to List"]</a>
-</div>
-
-@section Scripts {
-    @{await Html.RenderPartialAsync("_ValidationScriptsPartial");}
```
marcbeermann: We don't still need the validation scripts?

src/CollAction/Views/Projects/Find.cshtml
stewartmatheson: I looked at this page for a react proof of concept. Since all of it's HTML is due to be replaced anyway I left it out.

logophobia: I'd prefer the functionality stays in working condition in the master branch, even if it's due to be replaced with react.

logophobia: Unless that's significantly more work?

stewartmatheson: Why don't I just roll the template file back that way the page will stay functional. I can leave in the stubbed out react components in the frontend application and not load them in to the page. That way we can cut back over when Mina or I are ready to start work on the find project page.

stewartmatheson: Also would you mind if we migrated the install steps to the README of the application. Might help new users finding us through git who don;t have access to atlassian cloud.

stewartmatheson: Just rolled back the find template.

src/CollAction/Views/Projects/Create.cshtml
stewartmatheson: The server will still validate but it will be a bit heavy loading both jquery and the whole react bundle. There is a way to get both the client side react application as well as client side form validation based on the ASP.net view model pattern but I was going to do that post launch. Thoughts?

marcbeermann: Sounds like you've got it covered with a good plan ;)

logophobia:
Looks good, feel free to merge

stewartmatheson:
@logophobia I don't have access. Can you merge?

PR #63, opened 2017-02-13
CA-163, 170, 165 (Subtasks of CA-96)

marcbeermann:
ProjectsController.cs - Enabled banner image upload in the Start project page and Edit page.
+ ProjectBannerImageManager.cs
    - uploads banner images to the file system and updates the Project and ImageFile model accordingly.
    - uploaded banner images are stored to folder "wwwroot\usercontent\bannerimages"
+ SDImageMetaDataReader.cs
    - reads width and height info from image files
    - added new project depency CoreCompat.System.Drawing
+ Migration ProjectBannerImage.cs
    - Project.BannerImage column to index into ImageFile table.
+ ValidationAttributes
    - Validate file size, type and min image dimensions.


src/CollAction/Helpers/ProjectBannerImageManager.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace CollAction.Helpers
+{
+    public class ProjectBannerImageManager : IImageFileManager
+    {
+        public Project Project { get; set; }
+
+        private string WebFolder { get { return Path.Combine("usercontent", "bannerimages"); } }
+
+        // Don't trust the users _formFile.FileName to be unique, so generate our own unique one "banner_<Project.Id>.<FileType>"
```
marcbeermann: The comment doesn't match the code. I will change this comment to: ... own unique one "banner_<formatted-Project.Name>.<FileType>".

src/CollAction/Models/ProjectViewModels/CreateProjectViewModel.cs
```diff
@@ -58,5 +60,13 @@ public CreateProjectViewModel(SelectList categories, SelectList locations)
         [Required]
         [DisplayFormat(DataFormatString = "{0:dd/MM/yyyy}", ApplyFormatInEditMode = true)]
         public DateTime End { get; set; }
+
+        [Display(Name = "Banner image", Prompt = "1366x767px JPEG, GIF, PNG, BMP")]
+        [FileSize(1024000)] // 1MB
+        [FileType("jpg, jpeg, gif, png, bmp")]
+        [MinImageDimensions("1366x768")]
```
marcbeermann: As per CA-170 "For the header: maximum 1024 x 768 op 72 dpi" I will be changing the MinImageDimensionsAttribute class to MaxImageDimensions("1024x768").

src/CollAction/Models/ProjectViewModels/EditProjectViewModel.cs
```diff
@@ -60,5 +62,15 @@ public EditProjectViewModel(SelectList categories, SelectList locations)
         [Required]
         [DisplayFormat(DataFormatString = "{0:dd/MM/yyyy}", ApplyFormatInEditMode = true)]
         public DateTime End { get; set; }
+
+        [Display(Name = "Banner image", Prompt = "1366x767px JPEG, GIF, PNG, BMP")]
+        [FileSize(1024000)] // 1MB
+        [FileType("jpg, jpeg, gif, png, bmp")]
+        [MinImageDimensions("1366x768")]
```
marcbeermann: As per CA-170 "For the header: maximum 1024 x 768 op 72 dpi" I will be changing the MinImageDimensionsAttribute class to MaxImageDimensions("1024x768").

src/CollAction/Models/Project.cs
```diff
@@ -60,7 +60,11 @@ public class Project
         [Required]
         [MaxLength(2048)]
         public string CreatorComments { get; set; }
-        
+
+        public int? ImageFileId { get; set; }
```
logophobia: BannerImage(File)Id?

I'd imagine it's possible for more images to be associated with the project table.

src/CollAction/ValidationAttributes/MinImageDimensionsAttribute.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Linq;
+
+namespace CollAction.ValidationAttributes
+{
+    public class MinImageDimensionsAttribute : ValidationAttribute
+    {
+        private readonly int _minWidth;
+        private readonly int _minHeight;
+
+        public MinImageDimensionsAttribute(string widthByHeight)
```
logophobia: Why not two int parameters? Seems a bit akward taking a string, which is not type-safe, and then parsing it to two dimensions.

src/CollAction/ValidationAttributes/FileTypeAttribute.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.AspNetCore.Http;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.IO;
+using System.Linq;
+
+namespace CollAction.ValidationAttributes
+{
+    public class FileTypeAttribute : ValidationAttribute
```
logophobia: This seems like a decent idea, but is unused. Don't check in classes you don't use.

src/CollAction/ValidationAttributes/FileSizeAttribute.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Microsoft.AspNetCore.Http;
+using System.ComponentModel.DataAnnotations;
+
+namespace CollAction.ValidationAttributes
+{
+    public class FileSizeAttribute : ValidationAttribute
```
logophobia: This seems like a decent idea, but is unused. Don't check in classes you don't use.

src/CollAction/Helpers/SDImageMetaDataReader.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Collections.Generic;
+using System.Drawing;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace CollAction.Helpers
+{
+    // System.Drawing.Image implementation of the IImageMetaDataReader interface.
+    public class SDImageMetaDataReader : IImageMetaDataReader
```
logophobia: This feels a bit like premature abstraction. Not everything needs to be wrapped in interfaces and classes, KISS and YAGNI (keep-it-simple and you-aren't-to-need-it).

All abstraction has a price, an overhead wrt reading code, maintenance, refactoring. Since this literally only wraps 3 methods of the Image class, without adding anything, it seems rather superfluous. Also, if you create classes in the future, if you can, make them immutable (http://stackoverflow.com/questions/3769607/why-do-we-need-immutable-class), it's usually easier to program against immutable classes, harder to use them incorrectly.

src/CollAction/Helpers/ProjectBannerImageManager.cs
```diff
@@ -0,0 +1,68 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace CollAction.Helpers
+{
+    public class ProjectBannerImageManager : IImageFileManager
```
logophobia: It would probably be better to combine ProjectBannerImageManager, IImageFileManager and IFileManager in one class for now. This is probably another case of slight premature abstraction. Since we don't known if we're going to have multiple file managers, or multiple image file managers, it seems a bit much to use double inheritance, makes the code rather hard to follow. Maybe have some appropiate constructor parameters for the class, that make it generic to use (like begin able to customize the root directory where the images are stored). Also, https://en.wikipedia.org/wiki/Composition_over_inheritance.

src/CollAction/ValidationAttributes/FileTypeAttribute.cs
marcbeermann: This is used in EditProjectViewModel.cs line 68 and CreateProjectViewModel.cs line 66.

src/CollAction/ValidationAttributes/FileSizeAttribute.cs
marcbeermann: This is used in EditProjectViewModel.cs line 67 and CreateProjectViewModel.cs line 65.

src/CollAction/Controllers/ProjectsController.cs
```diff
@@ -214,6 +236,25 @@ public ViewResult StartInfo()
...
+                    WebFolder = Path.Combine("usercontent", "bannerimages")
+                };
+
+                if (project.BannerImage != null)
+                {
+                    manager.DeleteImageFile(project.BannerImage);
+                }
+
+                // Don't trust the users _formFile.FileName to be unique, so generate our own unique one "banner_<formatted-Project.Name>.<FileType>"
+                var uniqueFileName = (new Regex(@"[^A-Za-z0-9]+")).Replace("banner_" + project.Name, "_").Trim(new char[] { '_' });
```
logophobia: Maybe if we're doing this more often, add this property (NormalizedProjectName) to the model?

Also, this might be simpler: string uniqueFileName = Guid.NewGuid().ToString();

Either is fine.

src/CollAction/ValidationAttributes/FileTypeAttribute.cs
logophobia: My bad, shouldn't search for FileTypeAttribute, but FileType.

One small thing, if you make the constructor to be this:

 public FileTypeAttribute(params[] string types)
 {
     _types = types;
 }

Then you could use the attribute like this:
[FileType("jpg", "png")]

https://msdn.microsoft.com/nl-nl/library/w5zay9db.aspx

src/CollAction/Controllers/ProjectsController.cs
```diff
@@ -120,9 +126,24 @@ public ViewResult StartInfo()
...
                 End = createProjectViewModel.End,
-                Start = DateTime.UtcNow
+                Start = DateTime.UtcNow,
+                BannerImage = null 
             };
-            
+
+            if (createProjectViewModel.HasBannerImageUpload)
+            {
+                var manager = new ImageFileManager
```
logophobia: ImageFileManager probably needs a constructor. It's not a "pure data" class, and there are a couple of ways you could initialize it wrong.

A constructor would ensure people can't initialize the class wrong, and would take a context, a webroot and a webfolder.

src/CollAction/Helpers/ImageFileManager.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace CollAction.Helpers
+{
+    public class ImageFileManager
+    {
+        private IFormFile _formFile = null;
+        private ImageFile _imageFile = null;
```
logophobia: Unused private variable

src/CollAction/Helpers/ImageFileManager.cs
```diff
@@ -0,0 +1,98 @@
...
+        private string FileName { get { return String.Format("{0}{1}", _fileName, Path.GetExtension(_formFile.FileName).ToLower()); } }
+        
+        private string WebPath { get { return Path.Combine(WebFolder, FileName); } }
+
+        public async Task<ImageFile> UploadFormFile(IFormFile formFile, string fileName)
+        {
+            if (formFile == null) { return null; }
+
+            _formFile = formFile;
+            _fileName = fileName;
```
logophobia: It's probably not a good idea to have state in class that's only valid for one method call. Unnecessary state can be a source of bugs, and can make code harder to follow. Also, the state still exists afterwards while the FileManager keeps existing.

Prehaps a helper function would be easier, combined with passing formFile and fileName by parameter?
private string GetWebPath(string fileName, string extension) ...

logophobia:
Some pretty good improvements. I found some small bits that could be better, but good work.

logophobia:
Looks good, don't forget FileTypeAttribute.

logophobia:
Looks good, do you need me to merge it?

PR #64, opened 2017-02-16
CA_184 Fixed "logged off users cannot view the find or index page" bug

marcbeermann:
- also removed the edit link from the details page

### DIFF
--- a/src/CollAction/Controllers/ProjectsController.cs
+++ b/src/CollAction/Controllers/ProjectsController.cs
@@ -40,7 +40,7 @@ namespace CollAction.Controllers
             {
                 return View(new FindProjectViewModel
                 {
-                    OwnerId = (await _userManager.GetUserAsync(User)).Id,
+                    OwnerId = _userManager.GetUserId(User),
                     Projects = await DisplayProjectViewModel.GetViewModelsWhere(_context, p => p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted)
                 });
             }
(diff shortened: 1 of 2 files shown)


src/CollAction/Controllers/ProjectsController.cs
```diff
@@ -40,7 +40,7 @@ public ViewResult StartInfo()
             {
                 return View(new FindProjectViewModel
                 {
-                    OwnerId = (await _userManager.GetUserAsync(User)).Id,
+                    OwnerId = _userManager.GetUserId(User),
```
logophobia: removing "await" means the find page takes a bit more resources.

Have you tried:
* User == null ? null : (await _userManager.GetUserAsync(User)).Id
Or
* (await _userManager.GetUserAsync(User))?.Id

src/CollAction/Controllers/ProjectsController.cs
```diff
@@ -110,7 +110,7 @@ public ViewResult StartInfo()
 
             var project = new Project
             {
-                OwnerId = (await _userManager.GetUserAsync(User)).Id,
```
logophobia: Unrelated to your item, and again removes await, is there a reason here?

logophobia:
Looks good, if you tested it, feel free to merge

src/CollAction/Views/Projects/Details.cshtml
```diff
@@ -89,6 +89,5 @@
     </dl>
 </div>
 <div>
-    <a asp-action="Edit" asp-route-id="@Model.Project.Id">@Localizer["Edit"]</a> |
```
logophobia: I'm assuming this was added because it caused issues with users not logged in? Which is fine btw.

PR #65, opened 2017-02-21
CA-97, 162 Added a VideoLinks table and a reference to it from the Pr…

marcbeermann:
…oject table.

### DIFF
--- a/src/CollAction/Models/VideoLink.cs
+++ b/src/CollAction/Models/VideoLink.cs
@@ -1,0 +1,19 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace CollAction.Models
+{
+    public class VideoLink
+    {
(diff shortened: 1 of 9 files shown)


stewartmatheson:
Looks good to me

src/CollAction/Models/VideoLink.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace CollAction.Models
+{
+    public class VideoLink
+    {
+        [Key]
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        public int Id { get; set; }
+
+        [Required]
+        public string Link { get; set; }
```
logophobia: MaxLength 2083? See: http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

Some sort of maximum string length is a bit more efficient in a table

logophobia:
Looks good yeah.

PR #67, opened 2017-02-25
CA-182 Project Categories Endpoint

minayao:


src/CollAction/Controllers/ProjectsController.cs
```diff
@@ -274,5 +274,11 @@ public ViewResult StartInfo()
             await _context.SaveChangesAsync();
             return RedirectToAction("Index");
         }
+
+        [HttpGet]
+        public async Task<JsonResult> GetCategories()
```
stewartmatheson: Will this filter by category?

logophobia:
Looks good.

Unrelated, maybe we should try to agree on an api convention. "projects/categories" seems a bit arbitrary, maybe something REST like?

PR #68, opened 2017-03-02
1.1 upgrade (dotnet core and ef), integrating the build and docker compose

logophobia:
Ok, so this contains:

* Upgrade to dotnet core 1.1 and EF 1.1
* Integrating the frontend build automatically into the visual studio build
* A working docker compose build

In the end, I decided to keep the frontend folder in the csproj to have an easier build.

Before the c# build, npm install and npm run build (which calls webpack) is called.


docker-compose.vs.debug.yml
```diff
@@ -0,0 +1,20 @@
...
+  collaction:
+    image: collaction:dev
+    build:
+      args:
+        source: ${DOCKER_BUILD_SOURCE}
+    environment:
+      - DOTNET_USE_POLLING_FILE_WATCHER=1
+    volumes:
+      - ./CollAction:/app
+      - ~/.nuget/packages:/root/.nuget/packages:ro
```
stewartmatheson: is this ok under windows?

logophobia:
The ~ paths refer to paths in the base docker image. I don't think it requires any additional setup (I certainly never did any, and it works for me).

I merged your setup with the recommended docker-compose setup for asp.net core 1.1.

PR #71, opened 2017-03-05
CA-58

minayao:


CollAction/Helpers/ProjectService.cs
```diff
@@ -0,0 +1,62 @@
...
+using Microsoft.EntityFrameworkCore;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Threading.Tasks;
+
+namespace CollAction.Helpers
+{
+    public class ProjectService
```
logophobia: I think that the convention so far has been to add model related helper functions to the model classes themselves, see for example at the bottom of the project model. I'd rather keep that convention, easier to find code that way.

src/CollAction/Helpers/ProjectService.cs
```diff
@@ -0,0 +1,62 @@
...
+                UserId = userId,
+                ProjectId = projectId
+            };
+
+            try
+            {
+                _context.Add(participant);
+                await _context.SaveChangesAsync();
+            }
+            catch (Exception ex)
```
logophobia: If you don't use the "ex" variable, you can use:

catch (Exception)

logophobia:
Looks good

PR #74, opened 2017-03-20
AddMissingStartProjFields_CA_122

marcbeermann:
- Added in a 'Hashtag' field and a 'Start date' field that can be edited on the 'project create' page and 'project edit' page.
- Made 'Creator comments' a non-required field in line with CA_122 specs.
- Commented out the Locations SelectList code from the create and edit views and in the ProjectController (until we come up with a solution that doesn't load millions of locations into a select box).
- Moved the bannerimages folder back into the usercontent folder.


CollAction/Controllers/ProjectsController.cs
```diff
@@ -99,7 +99,8 @@ public ViewResult StartInfo()
             return View(new CreateProjectViewModel
             {
                 Categories = new SelectList(await _context.Categories.ToListAsync(), "Id", "Description"),
-                Locations = new SelectList(await _context.Locations.ToListAsync(), "Id", "Name", null)
+                // TODO: Add this back in when we've come up with a solution that doesn't involve returning the entire locations table!!!
```
logophobia: You can probably remove this entirely. The final solution will probably involve an API endpoint. We'll likely forget removing stuff like this when we're adding an API endpoint somewhere else.

CollAction/Views/Projects/Create.cshtml
```diff
@@ -76,17 +76,22 @@
                         @Html.DropDownList("CategoryId", Model.Categories)
                         <span asp-validation-for="CategoryId" class="text-danger" />
                     </div>
-                    <div class="form-group">
+                    @*<div class="form-group">
```
logophobia: Hmm, same thing as before. I'd rather we just clean-up the front-end code, and not comment it out. The former code won't be useful to the API endpoint (and if it is, it'll still be in git).

CollAction/Controllers/ProjectsController.cs
```diff
@@ -249,6 +264,8 @@ public ViewResult StartInfo()
                 project.BannerImage = await manager.UploadFormFile(editProjectViewModel.BannerImageUpload, Guid.NewGuid().ToString() /* unique filename */);
             }
 
+            await project.SetTags(_context, editProjectViewModel.Hashtag?.Split(';') ?? new string[0]);
```
logophobia: Doesn't matter now, but we should probably think up a better way to pass lists from the frontend to the backend (json, something buildin in asp.net core).

logophobia:
Looks good

PR #75, opened 2017-03-21
CA-181 Projects endpoint

minayao:


CollAction/Controllers/ProjectsController.cs
```diff
@@ -362,5 +362,35 @@ public ViewResult StartInfo()
...
+                    ProjectId = t.ProjectId,
+                    ProjectName = t.ProjectName,
+                    ProjectProposal = t.ProjectProposal,
+                    CategoryName = t.CategoryName,
+                    CategoryColorHex = t.CategoryColorHex,
+                    LocationName = t.LocationName,
+                    Target = t.Target,
+                    Participants = t.Participants,
+                    ProgressPercent = t.ProgressPercent,
+                    BannerImagePath = (t.BannerImage != null) ? Url.Content(t.BannerImage.Filepath.Replace('\\', '/')) : "https://placeholdit.imgix.net/~text?txtsize=33&txt=Project%20Image"
```
stewartmatheson: I have no idea about aspnet but should this banner selection code go in the ViewModel?

CollAction/Models/ProjectViewModels/DisplayTileProjectViewModel.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            if (remainingTime.TotalDays < 1)
+                RemainingTime = String.Format("{0} {1}", remainingTime.Hours, remainingTime.Hours == 1 ? "hour" : "hours");
+            else
+                RemainingTime = String.Format("{0} {1}", remainingTime.Days, remainingTime.Days == 1 ? "day" : "days");
+        }
+
+        public void setStatusTexts(ProjectStatus projectStatus, bool IsActive, bool isComingSoon, bool isClosed)
+        {
+            if (IsActive)
```
stewartmatheson: The only reason I would suggest that it stay on the server is to avoid the client having to manage translations however with that said I think the client will need to do that anyway so yeah if you feel it's better I'm happy to do that.

CollAction/Startup.cs
```diff
@@ -166,7 +166,12 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
                      "categories",
                      new { controller = "Projects", action = "GetCategories" }
                  );
-                
+
+                routes.MapRoute("GetTileProjects",
```
stewartmatheson: Yeah I would expect /Projects/Find just like we already have. The only difference would be the header that is passed in.

CollAction/Models/ProjectViewModels/DisplayTileProjectViewModel.cs
logophobia: If it stays on the backend, it should probably be localized. If Stewart is happy with either option, I'm happy as well.

CollAction/Startup.cs
```diff
@@ -163,12 +163,12 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
...
                 routes.MapRoute("getCategories",
-                     "categories",
+                     "api/categories",
                      new { controller = "Projects", action = "GetCategories" }
                  );
 
                 routes.MapRoute("GetTileProjects",
-                     "tileProjects",
+                     "api/projects/find",
                      new { controller = "Projects", action = "GetTileProjects" }
```
logophobia: Doesn't need to be changed now, but if our resty-API is supposed to be named "usage-independent" (just a general api, not dependent on what the frontend does with it), then I'd rather then backend functions are also named without mentioning tiled. The current Tile naming is going to be rather confusing if Stewart starts using it for different things.

CollAction/Startup.cs
```diff
@@ -163,10 +163,15 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
...
                 routes.MapRoute("getCategories",
-                     "categories",
+                     "api/categories",
                      new { controller = "Projects", action = "GetCategories" }
                  );
-                
+
+                routes.MapRoute("GetTileProjects",
+                     "api/projects/find",
+                     new { controller = "Projects", action = "GetTileProjects" }
```
logophobia: Doesn't need to be changed now, but if our resty-API is supposed to be named "usage-independent" (just a general api, not dependent on what the frontend does with it), then I'd rather then backend functions are also named without mentioning tiled. The current Tile naming is going to be rather confusing if Stewart starts using it for different things.

stewartmatheson:
Nothing jumps out at me as really bad. Feel free to merge if everyone else is happy.

CollAction/Models/ProjectViewModels/DisplayTileProjectViewModel.cs
```diff
@@ -0,0 +1,94 @@
+﻿using Microsoft.Extensions.Localization;
+using System;
+
+namespace CollAction.Models
+{
+    public class DisplayTileProjectViewModel
+    {
+        private readonly IStringLocalizer _localizer;
+
+        public DisplayTileProjectViewModel(IStringLocalizer localizer, DateTime projectStart, DateTime projectEnd, ProjectStatus projectStatus, bool IsActive, bool isComingSoon, bool isClosed)
```
logophobia: The new localization looks good

PR #76, opened 2017-03-23
RegistrationPage_CA_171

marcbeermann:
- Added a NewsletterSubscription field to the ApplicationUser model and a checkbox in the registration page (I've set it to be unchecked by default).
- Added a "Confirm email" input.
- Added a style sheet for admin pages to a frontend\admin folder.
- Moved the globals style sheet to the frontend\global folder.


CollAction/Models/ApplicationUser.cs
```diff
@@ -11,6 +11,7 @@ public class ApplicationUser : IdentityUser
         public string FirstName { get; set; }
         [MaxLength(250)]
         public string LastName { get; set; }
+        public bool NewsletterSubscription { get; set; }
```
logophobia: I was under the impression that we need newsletter subscriptions for both registered users, in that case we'd want to store the subscriptions in a separate NewsletterSubscription table. Check with Daniella?

CollAction/Controllers/AccountController.cs
```diff
@@ -114,7 +114,8 @@ public IActionResult Register(string returnUrl = null)
             ViewData["ReturnUrl"] = returnUrl;
             if (ModelState.IsValid)
             {
-                var user = new ApplicationUser { UserName = model.Email, Email = model.Email, NewsletterSubscription = model.NewsletterSubscription };
+                var user = new ApplicationUser { UserName = model.Email, Email = model.Email };
+                user.SetNewsletterSubscription(_context, model.Email, model.NewsletterSubscription);
```
marcbeermann: Missing the 'await' in front of user.SetNewletterSubscription().

CollAction/Models/ApplicationUser.cs
```diff
@@ -1,17 +1,47 @@
...
         public List<ProjectParticipant> Participates { get; set; }
+
+        public int? NewsletterSubscriptionId { get; set; }
+        [ForeignKey("NewsletterSubscriptionId")]
+        public NewsletterSubscription NewsletterSubscription { get; set; }
+
+        public async Task SetNewsletterSubscription(ApplicationDbContext context, string email, bool wantsSubscription)
+        {
+            NewsletterSubscription subscription = await context.NewsletterSubscriptions.SingleOrDefaultAsync(s => s.Email.Equals(email, StringComparison.Ordinal));
+            if (wantsSubscription)
```
logophobia: In both branches you change things, but in only one branch you call save changes

CollAction/Models/ApplicationUser.cs
```diff
@@ -1,17 +1,47 @@
...
+
         public List<ProjectParticipant> Participates { get; set; }
+
+        public int? NewsletterSubscriptionId { get; set; }
+        [ForeignKey("NewsletterSubscriptionId")]
+        public NewsletterSubscription NewsletterSubscription { get; set; }
+
+        public async Task SetNewsletterSubscription(ApplicationDbContext context, string email, bool wantsSubscription)
+        {
+            NewsletterSubscription subscription = await context.NewsletterSubscriptions.SingleOrDefaultAsync(s => s.Email.Equals(email, StringComparison.Ordinal));
```
logophobia: Make sure the ordinal string comparison is translated to sql, sql usually compares strings with a specific locale, if EF tries to table the query literally, it would be executed in memory.

stewartmatheson:
front-end code looks good to me.

CollAction/Models/ApplicationUser.cs
```diff
@@ -28,18 +28,14 @@ public class ApplicationUser : IdentityUser
 
         public async Task SetNewsletterSubscription(ApplicationDbContext context, string email, bool wantsSubscription)
         {
-            NewsletterSubscription subscription = await context.NewsletterSubscriptions.SingleOrDefaultAsync(s => s.Email.Equals(email, StringComparison.Ordinal));
+            NewsletterSubscription subscription = await context.NewsletterSubscriptions.SingleOrDefaultAsync(s => s.Email.ToLower() == email.ToLower());
```
logophobia: This won't use the index you created on email. We should change the type to citext to ensure you don't need to lower, or accept that we can't do this efficiently for now. See: https://www.postgresql.org/docs/9.1/static/citext.html

Sorry, should've realized this earlier.

logophobia:
Looks good, no issues as far as I can see.

marcbeermann:
This pull request is superseded by the NewsletterSubscriptions PR and is thus being removed.

CollAction/Models/AccountViewModels/RegisterViewModel.cs
```diff
@@ -29,5 +33,8 @@ public class RegisterViewModel
 
         [Display(Name = "Last name")]
         public string LastName { get; set; }
+
+        [Display(Name = "I would like to receive an update from CollAction every once in a while - don't worry, we like spam as little as you do! \U0001F642")]
```
logophobia: Doesn't really matter, but you don't need to escape the unicode, you can just paste 🙂 in the cs file.

PR #78, opened 2017-04-02
Project listing page

stewartmatheson:
1. Added the project search page with requests working back from the server
2. Added about us page
3. Added responsive styles
4. Added mobile first about page.


CollAction/Services/ProjectService.cs
```diff
@@ -77,7 +77,7 @@ public ProjectService(ApplicationDbContext context, IStringLocalizer<IProjectSer
                     Target = p.Target,
                     Participants = p.Participants.Count()
                 })
-               .ToListAsync();
+               .ToList();
```
logophobia: Is this on purpose? Making it synchronous ensures a thread is occupied while it queries the database.

logophobia:
The rest looks good, not that I'm a good judge of react code.

CollAction/Services/ProjectService.cs
stewartmatheson: Mistake. Good catch. Will fix tonight.

PR #80, opened 2017-04-10
forbid visual studio from compiling the typescript files itself

logophobia:
Fix visual studio trying to build the typescript files with its new typescript integration.

In some future, we might want to use that integration, might be nice for development, for now, this is a quick fix.


stewartmatheson:
Looks good to me. Merge away.

PR #89, opened 2017-04-24
Added two packages to provide Core compatibilitys with SystemDrawing …

marcbeermann:
…on both linux and osx platforms.


logophobia:
Tried this locally, couldn't get it working inside docker. I made a solution with ImageSharp here: https://github.com/CollActionteam/CollAction/pull/92, since I'm assuming you're a bit busy.

PR #93, opened 2017-04-28
CA-188 status endpoint, update project endpoint with status

minayao:


CollAction/Controllers/ProjectsController.cs
```diff
@@ -364,12 +365,35 @@ public ViewResult StartInfo()
...
         [HttpGet]
-        public async Task<JsonResult> GetTileProjects(int? categoryId, int? locationId)
+        public async Task<JsonResult> GetTileProjects(int? categoryId, int? locationId, string statusId)
         {
-            var displayTileProjectViewModels = await _service.GetTileProjects(Url, p => p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted
+            Expression<Func<Project, bool>> projectExpression = (p => p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted
             && (categoryId != null ? p.CategoryId == categoryId : true) && (locationId != null ? p.LocationId == locationId : true));
+
+            Expression<Func<Project, bool>> statusExpression;
+            switch (String.IsNullOrEmpty(statusId) ? String.Empty : statusId.ToLower())
```
logophobia: The ternary statement here is handled by the "default" case.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -364,12 +365,35 @@ public ViewResult StartInfo()
...
+            {
+                case "open": statusExpression = (p => p.IsActive == true); break;
```
logophobia: You're using a lot of unmapped properties here. It's best not to search for unmapped properties, those don't get optimized to sql. Very inefficient, means the whole project table is selected, and then filtered client-side.

Just search for the implementation, like:
        public bool IsClosed
            => Status == ProjectStatus.Running && End < DateTime.UtcNow;

CollAction/Controllers/ProjectsController.cs
```diff
@@ -364,12 +365,35 @@ public ViewResult StartInfo()
         }
 
         [HttpGet]
-        public async Task<JsonResult> GetTileProjects(int? categoryId, int? locationId)
+        public async Task<JsonResult> GetTileProjects(int? categoryId, int? locationId, string statusId)
         {
-            var displayTileProjectViewModels = await _service.GetTileProjects(Url, p => p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted
+            Expression<Func<Project, bool>> projectExpression = (p => p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted
```
logophobia: I'm not sure if ternary statements get optimized to sql well. Maybe check in a query tracer if it gets optimized?

If not, I'd pull the parameter check outside the query body, like:

if (categoryId != null)
    projectExpression = projectExpression.AndAlso..

CollAction/Controllers/ProjectsController.cs
```diff
@@ -364,12 +365,35 @@ public ViewResult StartInfo()
         }
 
         [HttpGet]
-        public async Task<JsonResult> GetTileProjects(int? categoryId, int? locationId)
+        public async Task<JsonResult> GetTileProjects(int? categoryId, int? locationId, string statusId)
```
logophobia: locationId is still not needed as a parameter, location-wise search will probably take the form of, longitude/latitude/radius.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -364,12 +365,35 @@ public ViewResult StartInfo()
         }
 
         [HttpGet]
-        public async Task<JsonResult> GetTileProjects(int? categoryId, int? locationId)
+        public async Task<JsonResult> GetTileProjects(int? categoryId, string statusId)
         {
-            var displayTileProjectViewModels = await _service.GetTileProjects(Url, p => p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted
-            && (categoryId != null ? p.CategoryId == categoryId : true) && (locationId != null ? p.LocationId == locationId : true));
+            Expression<Func<Project, bool>> projectExpression = (p => p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted
+            && (categoryId != null ? p.CategoryId == categoryId : true));
```
logophobia: Did you check if the category filter optimizes to sql? That's the only one left I'm unsure about.

For the rest it looks good.

minayao: Ya, the sql looked fine. Category's only included in the where clause when not null.

logophobia: Ok, feel free to merge then

PR #103, opened 2017-05-04
Start and End date validation CA_207 and CA_197:

marcbeermann:
- added a validation attribute 'DateWithinMonths' to ensure that the startdate is at least within a year of today's date.
- added a validation attribute 'DateWithinMonths' to ensure that the enddate is within a year of the start date.
- updated CreatProject/EditProject/ManageProjectViewModels accordingly.
- moved the YouTubeLinkAttribute.cs file up a level.
- fixed broken links to edit project from the Manage index page.


CollAction/ValidationAttributes/DateWithinMonthsAttribute.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Reflection;
+
+namespace CollAction.ValidationAttributes
+{
+    public class DateWithinMonthsAttribute : ValidationAttribute
+    {
+        private readonly int _months;
+
+        public String OfDateProperty { get; set; }
```
logophobia: I'd use two validation attributes for this. For comparing to another date, against comparing to the current date. They're different things. Separation of concerns.

CollAction/ValidationAttributes/DateWithinMonthsAttribute.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public DateWithinMonthsAttribute(int months)
```
logophobia: To indicate a difference in dates, it's better and more general to use a timespan: https://msdn.microsoft.com/en-us/library/system.timespan(v=vs.110).aspx

marcbeermann: Passing a TimeSpan as an attribute parameter is not possible (only primitive types can be passed as attribute parameters).

marcbeermann: I could make it more generic by passing in years, months and days as attribute parameters. I could also pass in the parameters necessary to make up a TimeSpan hours, minutes, seconds (but I think that is too fine a granularity for our purposes). Also, if we can pass in months rather than a fixed time span that gets around odd leap year behavior (off by one day).

logophobia: Ah, I see your point about leap years. Sure, go ahead and cell in.

There are workarounds for the primitive type issue btw: https://msdn.microsoft.com/en-us/library/system.configuration.timespanvalidatorattribute(v=vs.110).aspx .

logophobia:
Go ahead and merge it

PR #116, opened 2017-05-12
get currently selected statusId

minayao:


logophobia:
That looks like a proper fix, merge away

PR #120, opened 2017-05-13
mobile nav open too far to the right on tablets

minayao:


logophobia:
If it works, go ahead and merge

PR #129, opened 2017-05-27
amazon ses instead of sendgrid (CA-222 and CA-269)

logophobia:


CollAction/Startup.cs
```diff
@@ -84,9 +84,10 @@ public void ConfigureServices(IServiceCollection services)
             services.AddTransient<ISmsSender, AuthMessageSender>();
             services.Configure<AuthMessageSenderOptions>(options =>
             {
-                options.FromName = Configuration["FromName"];
                 options.FromAddress = Configuration["FromAddress"];
-                options.SendGridKey = Configuration["SendGridKey"];
+                options.Region = RegionEndpoint.EnumerableAllRegions.First(r => r.SystemName == Configuration["SesRegion"]);
+                options.SesAwsAccessKey = Configuration["SesAwsAccessKey"];
```
stewartmatheson: Does this get set from the environment?

env_collaction.template
```diff
@@ -3,10 +3,10 @@ DbHost=db
 DbUser=postgres
 DbPort=5432
 Db=CollActionDb
-FromName=
 FromAddress=your@email
-SendGridUser=
-SendGridKey=
+SesAwsAccessKey=
+SesAwsAccessKeyId=
+SesRegion=
```
marcbeermann: Maybe update the 'Getting CollAction running locally' with the Ses configurations.

CollAction/Startup.cs
logophobia: Either the environment, a config file or user secrets (in debug).

env_collaction.template
logophobia: I will.

PR #135, opened 2017-06-13
CA_228_ManageProjectsShowExistingBannerImage

marcbeermann:
- Made the Manage Projects page show the existing banner image and descriptive image.
- Fixed a bug where you were not able to update the image descriptions if you didn't upload an image at the same time.
- Added validation messages to inform the user that the image description is only saved if they actually upload an image.
- Updated the ProjectsController to resolve the same two points above.
- Made the ImageFileManager able to deal with both the mac/unix style file paths and the windows style file paths without breaking.


logophobia:
Question, was this also tested in docker? Because, if you're making this windows compatible (which is a bit unnecessary since we're always hosting in docker), it'd be a really good idea to test it properly in an environment similar to where it'll be hosted.

CollAction/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -109,16 +109,16 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<string>("Description")
                         .IsRequired();
 
-                    b.Property<string>("Filepath")
```
logophobia: This seems to be the only migration change? I think you forgot to undo this.

PR #140, opened 2017-06-21
CA-274-FooterLinksDoNotWorkFix

marcbeermann:
- Got all the links working appropriately bar one (the 'press' link doesn't have a corresponding press section in the about page).
- Also added a small javascript to the about page in order to fix the navigation bar offsetting the page incorrectly when navigating via a link to a page fragment.

### DIFF
--- a/CollAction/Views/Home/About.cshtml
+++ b/CollAction/Views/Home/About.cshtml
@@ -139,3 +139,21 @@
     </div>
 
 </div>
+
+<script type="text/javascript">
+    var shiftWindowToCorrectForNavBarHeight = function ()
+    {
+        var navBarHeight = document.querySelector('.nav-row').offsetHeight;
+        scrollBy(0, -navBarHeight);
+    };
+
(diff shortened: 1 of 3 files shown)


CollAction/Views/Home/About.cshtml
```diff
@@ -139,3 +139,21 @@
     </div>
 
 </div>
+
+<script type="text/javascript">
+    var shiftWindowToCorrectForNavBarHeight = function ()
+    {
+        var navBarHeight = document.querySelector('.nav-row').offsetHeight;
```
logophobia: Not sure about this, maybe stew knows a cleaner way?

stewartmatheson: Yeah Ideally we should not have to use JS to change heights. I'm also not that keen in having too much javascript outside the frontend project. Is there a way to do this using CSS?

logophobia: @stewartmatheson We should probably either:
* Give poor marc an alternative solution
* Or merge this and fix later

Since you're the front-end expert, I'll leave that to you.

PR #143, opened 2017-07-16
Ca 278

minayao:


CollAction/Models/ProjectViewModels/DisplayProjectViewModel.cs
```diff
@@ -37,6 +37,26 @@ private string YouTubeId
...
             }
         }
+        
+        [DataType(DataType.Date)]
+        [Display(Name = "Start date")]
+        public DateTime Start 
+        {
+            get
+            {
+                return Project.End;
```
marcbeermann: Is this supposed to be return Project.Start ?

PR #151, opened 2017-07-30
CA-293 fixes

logophobia:
The changes are up on the testserver.


logophobia:
Acceptance tests were a success

PR #157, opened 2017-08-27
Fix build issues due to release new visual studio / dotnet core 2.0

logophobia:


CollAction/CollAction.csproj
```diff
@@ -20,27 +21,7 @@ npm run build</PreBuildEvent>
   </PropertyGroup>
 
   <ItemGroup>
-    <Content Include="wwwroot\images\team\Brian Russell.png" />
```
stewartmatheson: These where causing the build to break for me. I have already deleted them from my working copy.

CollAction/Dockerfile
```diff
@@ -1,4 +1,4 @@
-﻿FROM microsoft/aspnetcore:1.1.1
+﻿FROM microsoft/aspnetcore:1.1.2
```
stewartmatheson: So we still need to build first on the host OS which is causing some issues. Is there an easy way to get a dev server working within docker too?

docker-compose.vs.debug.yml
```diff
@@ -1,20 +0,0 @@
-version: '2'
```
stewartmatheson: How come we dont need this any more?

docker-compose.dcproj
```diff
@@ -5,17 +5,13 @@
     <DockerLaunchBrowser>True</DockerLaunchBrowser>
```
stewartmatheson: Is this file just for visual studio? Is there a way to run this from the command line tools?

CollAction/Dockerfile
logophobia: No easy, but maybe we should make a build docker image

docker-compose.vs.debug.yml
logophobia: It's been replaced by version: '3'

The whole update process is something visual studio does automatically.

docker-compose.dcproj
logophobia: It's a visual studio project to run that enables you to run it in docker automatically.

PR #161, opened 2017-09-03
Docker dev server

stewartmatheson:


CollAction/CollAction.csproj
```diff
@@ -1,47 +1,21 @@
...
-
   <PropertyGroup>
     <TargetFramework>netcoreapp1.1</TargetFramework>
     <TypeScriptCompileBlocked>true</TypeScriptCompileBlocked>
     <PackageTargetFallback>$(PackageTargetFallback);portable-net45+win8+wp8+wpa81;</PackageTargetFallback>
-    <DisableImplicitFrameworkReference>true</DisableImplicitFrameworkReference>  
+    <DisableImplicitFrameworkReference>true</DisableImplicitFrameworkReference>
     <UserSecretsId>aspnet-CollAction-29a5d7b8-3530-4482-9ed6-572bd6c178fa</UserSecretsId>
     <DockerComposeProjectPath>..\docker-compose.dcproj</DockerComposeProjectPath>
-    <PreBuildEvent>npm install &amp;&amp; ^
```
logophobia: This means we don't get a rebuild frontend in visual studio. I'd prefer not to break other ways to test.

CollAction/Dockerfile
```diff
@@ -1,7 +1,8 @@
-﻿FROM microsoft/aspnetcore:1.1.1
-ARG source
+﻿FROM microsoft/dotnet:1.1.2-sdk
```
logophobia: I think it's probably best if we use a *separate* docker image to develop, build and test. This is also the docker image used for the production server, it doesn't really need the sdk. Also, the aspnetcore probably has a few good tweaks in it for a production server.

logophobia:
You should probably resolve the conflicts with the main branch. I think a lot was cherry-picked or manually copied over?

docker-compose.yml
```diff
@@ -19,13 +19,23 @@ services:
...
     volumes:
       - images:/app/wwwroot/usercontent
+      - "./CollAction:/app"
     environment:
-      - ASPNETCORE_ENVIRONMENT=Release
+      - ASPNETCORE_ENVIRONMENT=Development
       - ASPNETCORE_URLS=http://*:5000
+    command: [ "run" ]
+
+  collaction-frontend:
```
logophobia: Question, why do we have two entries for both the frontend and the backend? They're run in the same container right?

stewartmatheson: There is a process watching the frontend for rebuilds and a process running the backend. We should have a container for each process.

CollAction/Dockerfile
stewartmatheson: Fair point. What about a Dockerfile and a Dockerfile.local? The Dockerfile.local can run local dev and Dockerfile can remain where it is for prod.

CollAction/CollAction.csproj
stewartmatheson: Can PreBuildEvents only ben run for a given environment? Is there anyway we can do this just for production builds as we should not need this anymore in dev.

CollAction/Dockerfile
logophobia: Dockerfile and Dockerfile.dev (or test)

stewartmatheson: yep happy to do that. Will make the change.

CollAction/CollAction.csproj
stewartmatheson: @logophobia Hey Tim any thoughts on this one?

logophobia: Couple of options:
* Make it conditional so that you can disable with an evironment variable
* Make a separate csproj with that task disabled
* Disable it on build (not sure if that's possible with dotnet build)

https://stackoverflow.com/questions/1731904/can-you-prevent-msbuild-exe-from-running-build-events

stewartmatheson: I like the environment variable idea. I can set it within the docker compose file. Will try and have a look at this one some time today.

logophobia:
We'll be doing this slightly different. ASP.net core 2.1 has a template for this.

PR #165, opened 2017-10-01
CA-257 - Allow deleting images

logophobia:


logophobia:
Was acceptance tested

PR #168, opened 2017-10-18
Ca 190 homepage design feedback

ho23a:


CollAction/Views/Projects/Create.cshtml
```diff
@@ -21,17 +21,17 @@
                     </div>
```
logophobia: Don't think this is part of CA-190, shouldn't be in the changeset

logophobia:
@ljwboschker Could you review this as well? CA-190 is quite.. large. Brought things up to date, but still..

ljwboschker:
Moved the branch to the collaction git repo so it works with the new CI/CD environment.

PR #178, opened 2018-01-04
Ca 361 friesland startproject form

ho23a:
Set Category=Friesland for all new Friesland projects

### DIFF
--- a/CollAction/Controllers/ProjectsController.cs
+++ b/CollAction/Controllers/ProjectsController.cs
@@ -136,10 +136,9 @@ namespace CollAction.Controllers
                 Goal = model.Goal,
                 Proposal = model.Proposal,
                 CreatorComments = model.CreatorComments,
-                CategoryId = _context
+                CategoryId = (await _context
                     .Categories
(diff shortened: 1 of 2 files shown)


CollAction/Controllers/ProjectsController.cs
```diff
@@ -136,7 +136,10 @@ public ViewResult StartInfo()
                 Goal = model.Goal,
                 Proposal = model.Proposal,
                 CreatorComments = model.CreatorComments,
-                CategoryId = model.CategoryId,
+                CategoryId = _context
+                    .Categories
+                    .FirstOrDefault(c => c.Name == "Friesland")
```
logophobia: * SingleAsync() is probably the method that makes the most sense here:
  + Single because you want it to return exactly one record (https://stackoverflow.com/questions/3485317/entity-framework-4-single-vs-first-vs-firstordefault)
  + Async because that's more efficient for IO. Doesn't matter much here, but it's a good habit to use async to access the database

ho23a:
@logophobia noted and updated the code. Please take a look, thank you!

PR #182, opened 2018-01-07
Ca 348 friesland startproject form

ho23a:
- Change text to Dutch
- Remove Category label and box
- Switch bottom button color to blue #0168b3


CollAction/.gitignore
```diff
@@ -1,3 +1,5 @@
 wwwroot/js/bundle.js
 node_modules
-log*
\ No newline at end of file
+log*
+
+Views/Projects/Participate.cshtml
```
logophobia: Why exclude this?

ho23a: It is an incomplete file that I accidentally committed and pushed earlier.

logophobia: Could you undo it? Because it'll be merged to the friesland branch as it is now?

PR #185, opened 2018-01-13
CA-359-Friesland-project-tile

marcbeermann:


CollAction/Frontend/app/project/ProjectList.tsx
```diff
@@ -44,19 +45,19 @@ class ProjectThumb extends React.Component<IProject, null> {
...
             </div>
             <div>
               <div className="value">{this.props.target}</div>
-              <div className="label" >Target</div>
+              <div className="label">Target</div>
             </div>
           </div>
           <div className="project-thumb-button">
-            <a href={link} style={{backgroundColor: "#" + this.props.categoryColorHex}}>Read More</a>
+            <a href={link} style={{ backgroundColor: "#0168b3" }}>Lees meer</a>
```
ljwboschker: Isn't this backgroundColor part of the category in the database? If this webapp only shows projects in the Friesland-category, and that category has color #0168b3, then you don't need this hard-coded value, right?

PR #191, opened 2018-01-30
Social Media Logins and ASP .NET CORE upgrade

logophobia:
- Contains a migration to asp net core 2
- Added logins for google/facebook/twitter
- Revised structure of some controllers to resemble the way the "reference" application does things in asp net core 2
- Remove some unused code surrounding accounts. Probably not a good idea to leave active account related code that can't be reached/tested, security/efficiency wise

This will merge into master, I'll make a separate PR that merges into the friesland branch w/ the correct translations and layout.


CollAction/Program.cs
```diff
@@ -3,22 +3,24 @@
...
 using System.Linq;
 using System.Threading.Tasks;
+using CollAction;
+using Microsoft.AspNetCore;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
 
-namespace CollAction
+namespace WebApplication2
```
ljwboschker: Strange namespace...

CollAction/Startup.cs
```diff
@@ -67,6 +67,23 @@ public void ConfigureServices(IServiceCollection services)
                 options.Password.RequiredLength = 8;
             });
 
+            services.AddAuthentication()
+                    .AddFacebook(options =>
+                    {
+                        options.AppId = Configuration["Authentication:Facebook:AppId"];
```
ljwboschker: Where are these values coming from? I couldn't find them in the values. Should they be in env_collaction?

ljwboschker: Also, I think the keys, and the accounts you can use to manage them, should be stored in the password-tool.

CollAction/Views/Account/ExternalLogin.cshtml
```diff
@@ -0,0 +1,62 @@
...
+<div id="login-starting">
+    <div class="container">
+        <div class="row">
+            <div class="col-xs-12">
+                <h1>@ViewData["Title"]</h1>
+                <h4>Associate your @ViewData["LoginProvider"] account.</h4>
+                <p class="text-info">
+                    You've successfully authenticated with <strong>@ViewData["LoginProvider"]</strong>.
+                    Please enter an email address for this site below and click the Register button to finish
+                    logging in.
```
ljwboschker: The text could be less technical, more focussed on the user. Perhaps we can ask UX.

CollAction/Program.cs
```diff
@@ -3,13 +3,12 @@
...
 using System.Linq;
 using System.Threading.Tasks;
-using CollAction;
 using Microsoft.AspNetCore;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 
-namespace WebApplication2
+namespace CollAction
```
ljwboschker: Ah. Forget my previous comment about this :-)

CollAction/Program.cs
logophobia: This was corrected in a later commit. You can look at the "Changes" tab for the effective changes.

CollAction/Startup.cs
logophobia: They are stored in the same place as all the other keys, safely. Either in aws configuration, or the password tool.

But you're right, I should add them to the env_collaction.template as well, for easier configuration.

CollAction/Views/Account/ExternalLogin.cshtml
logophobia: I'll ask in slack

logophobia:
First I'll merge the english version with master (it's based on master since it's been in development so long), then I'll merge a dutch version with the friesland branch. Easier that way.

Also, changed the text and added the configuration options to the template.

PR #198, opened 2018-03-14
Ca 363 updating existing end dates

xprmnts:
1. Added Migration to Update existing End Dates by 23:59:59
2. Updated AdminController & ProjectController values for End Date Property to append 23:59:59
3. Added FAQ based on Ron's Comments in JIRA
4. Considered adding UTC on the Front End to Render "End Date (UTC)" but since there is no time being displayed it did not look right, so I decided not to make that change at the moment.


xprmnts:
Added .Date in both the Project and Admin Controllers... to ensure time addition doesn't drift

PR #199, opened 2018-03-16
Ca 372

ljwboschker:
Refactored the project embed widget so it uses the same code as the in-site project overview page. Basically:
* back-end API call now accepts an optional projectId query-parameter
* Embed.cshtml activates the FindProject react-component
* Add projectId property to FindProject react-component. If it's present, it will add that to the API request


ljwboschker:
Trying to re-open the pull request...

PR #200, opened 2018-03-16
Ca 372

logophobia:
Resubmitted @ljwboschker 's pull request, after it was merge a bit too soon.

ljwboschker:
Refactored the project embed widget so it uses the same code as the in-site project overview page. Basically:

* back-end API call now accepts an optional projectId query-parameter
* Embed.cshtml activates the FindProject react-component
* Add projectId property to FindProject react-component. If it's present, it will add that to the API request


CollAction/Controllers/ProjectsController.cs
```diff
@@ -464,21 +459,25 @@ public ViewResult StartInfo()
         }
 
         [HttpGet]
-        public async Task<JsonResult> GetTileProjects(int? categoryId, int? statusId)
+        public async Task<JsonResult> GetTileProjects(int? projectId)
```
logophobia: This API is used for other things. Normal project search is by category and status... so I'm pretty sure this breaks things. We should probably get a different API endpoint for this.

CollAction/Frontend/app/project/FindProject.tsx
```diff
@@ -4,6 +4,7 @@ import * as React from "react";
 import { IProject } from "./ProjectList";
 
 interface IFindProjectProps {
+  projectId?: string;
```
logophobia: Project id is an integer.

CollAction/Frontend/app/project/index.tsx
```diff
@@ -20,4 +20,13 @@ renderComponentIf(
...
\ No newline at end of file
+);
+
+renderComponentIf(
+  <FindProject controller={false} projectId={getQueryString("projectId") } />,
+  document.getElementById("embedded-projects-container")
+);
+
+function getQueryString(key) {
+  return decodeURIComponent(window.location.search.replace(new RegExp("^(?:.*[&\\?]" + encodeURIComponent(key).replace(/[\.\+\*]/g, "\\$&") + "(?:\\=([^&]*))?)?.*$", "i"), "$1"));
```
logophobia: This seems a .. very complicated way of doing things. Are you sure this can't be simplified a bit?

CollAction/Controllers/ProjectsController.cs
ljwboschker: Restored original function and added a new one to get just one project.

CollAction/Frontend/app/project/index.tsx
ljwboschker: There is; URLSearchParams. Unfortunately, not supported by all browsers (https://caniuse.com/#search=URLSearchParams). Got this one from StackOverflow.

CollAction/Frontend/app/project/FindProject.tsx
```diff
@@ -36,7 +37,16 @@ export default class FindProject extends React.Component<IFindProjectProps, IFin
 
     // Fetch projects with out filters set
     const getUrl: () => string = () => {
-      return "/api/projects/find";
+      let url = "/api/projects";
```
logophobia: When testing, we should ensure that the /find page still works. This shouldn't break it, but it does touch its code.

In the future, if these components start drifting apart in functionality, we should split them apart.

PR #213, opened 2018-04-13
Ca 267

ljwboschker:
I wrapped the rich text editor in react-quill, to make it into a React-component. This way the framework takes care of all lifecycle and updating events.

I also opted for storing the HTML that the component outputs. That save a lot of code and works pretty smooth. 

The input is sanitised (the parts that are rendered using @Html.Raw) using a whitelist of tags before it is stored (in the ProjectsController and the AdminController). That should prevent XSS. Did I forget any places where projects can be stored?


CollAction/CollAction.csproj
```diff
@@ -44,4 +40,7 @@ npm run build</PreBuildEvent>
   <ItemGroup>
     <Folder Include="wwwroot\js\" />
   </ItemGroup>
+  <ItemGroup>
+    <TypeScriptCompile Include="Frontend\app\global\RichTextEditor.tsx" />
```
logophobia: I thought we had our own frontend "builder" (through webpack, triggered through npm build) which takes care of the typescript. Had to explicitly disable the c# typescript stuff to stop it interfering. Why are we using the buildin typescript functionality here?

CollAction/Controllers/ProjectsController.cs
```diff
@@ -128,13 +129,15 @@ public ViewResult StartInfo()
                 return View(model);
             }
 
+            // TODO: check incoming HTML for invalid tags
```
logophobia: I think that's done here, looking at the code below. Bit confusing to leave "done" stuff as TODO comment.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -163,7 +166,7 @@ public ViewResult StartInfo()
                 "Hi!<br>" +
                 "<br>" +
                 "Thanks for submitting a project on www.collaction.org!<br>" +
-                "The CollAction Team will review your project as soon as possible � if it meets all the criteria we�ll publish the project on the website and will let you know, so you can start promoting it! If we have any additional questions or comments, we�ll reach out to you by email.<br>" +
+                "The CollAction Team will review your project as soon as possible � if it meets all the criteria we�ll publish the project on the website and will let you know, so you can start promoting it! If we have any additional questions or comments, we�ll reach out to you by email.<br>" +
```
logophobia: I'm getting invalid unicode characters here in this change? Are you using a mac? We had issues like that before. Undo?

CollAction/Controllers/ProjectsController.cs
```diff
@@ -415,7 +418,7 @@ public ViewResult StartInfo()
                     "Hi!<br><br>" +
                     "Thank you for participating in a CollAction project!<br><br>" +
                     "In crowdacting, we only act collectively when we meet the target before the deadline, so please feel very welcome to share this project on social media through the social media buttons on the project page!<br><br>" +
-                    "We�ll keep you updated on the project. Also feel free to Like us on <a href=\"https://www.facebook.com/collaction.org/\">Facebook</a> to stay up to date on everything CollAction!<br><br>" +
+                    "We�ll keep you updated on the project. Also feel free to Like us on <a href=\"https://www.facebook.com/collaction.org/\">Facebook</a> to stay up to date on everything CollAction!<br><br>" +
```
logophobia: I'm getting invalid unicode characters here in this change? Are you using a mac? We had issues like that before. Undo?

CollAction/Models/ProjectViewModels/CreateProjectViewModel.cs
```diff
@@ -34,16 +34,18 @@ public CreateProjectViewModel(SelectList categories, SelectList locations)
 
         [Required(ErrorMessage = "Give a succinct description of the issues your project is designed to address")]
         [Display(Name = "Short description", Prompt = "E.g Reduce plastic waste and save our oceans!")]
-        [MaxLength(1000)]
+        [MaxLength(10000)]
         public string Description { get; set; }
+        //public string DescriptionHtml { get; set; }
```
logophobia: Don't leave commented out code in the project

CollAction/Models/ProjectViewModels/CreateProjectViewModel.cs
```diff
@@ -34,16 +34,18 @@ public CreateProjectViewModel(SelectList categories, SelectList locations)
...
 
         [Required(ErrorMessage = "Describe what you hope to have achieved upon successful completion of your project")]
         [Display(Name = "Goal/Impact", Prompt = "Max 1000 characters")]
-        [MaxLength(1000)]
+        [MaxLength(10000)]
         public string Goal { get; set; }
+        //public string GoalHtml { get; set; }
```
logophobia: Don't leave commented out code in the project

CollAction/Migrations/20180329050159_ProjectRichText.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
logophobia: There're multiple migrations for this pull request. Could you compress them into one migration? Adding and then removing columns might result in a suboptimal disk-structure for the tables with some database systems.

CollAction/Helpers/InputSanitizer.cs
```diff
@@ -0,0 +1,20 @@
...
+        {
+            var saniziter = new HtmlSanitizer(
+                allowedTags: new[] { "p", "br", "strong", "em", "i", "u", "a", "ol", "ul", "li" },
+                allowedAttributes: new[] { "href", "target" },
+                allowedCssClasses: new string[] {},
+                allowedCssProperties: new string[] {}                    
+            );
+
+            return saniziter.Sanitize(input);
+
```
logophobia: Unnecessary empty line. Also, might be cleaner with a validation attribute: https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-2.1

logophobia: Also, the sanitizing might be a bit simplistic. For example, I think this will still get through:

<a href="javascript:alert(8007)">Click me</a>

You need to verify the contents of those attributes as well (for example, is the href a valid url with protocol http or https).

PR #217, opened 2018-04-21
EDIT- project name in urls CA289

dominicusplatus:
CA289

### DIFF
--- a/CollAction/Models/ProjectViewModels/CommitProjectViewModel.cs
+++ b/CollAction/Models/ProjectViewModels/CommitProjectViewModel.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
+using System.Net;
 
 namespace CollAction.Models
 {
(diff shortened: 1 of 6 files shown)


CollAction/Models/ProjectViewModels/CommitProjectViewModel.cs
```diff
@@ -15,6 +15,6 @@ public class CommitProjectViewModel
 
         public bool IsActive { get; set; }
 
-        public string ProjectLink => $"/Projects/Details/{ProjectId}";
+        public string ProjectLink => $"/Projects/{ProjectName}/Details";
```
logophobia: This isn't used

dominicusplatus: Yes but it is still in code so should be updated in case it is used in future, and we would need a story to cleanup unused things.

logophobia: Probably best to just clean it up. It's ok to clean stuff up related to the PR when you're working on it anyway, the functionality is tested anyway. As long as you limit the refactorings to the ticket.

dominicusplatus:
http://ec2-34-240-196-84.eu-west-1.compute.amazonaws.com:8001/

dominicusplatus:
duplicate https://github.com/CollActionteam/CollAction/pull/223

PR #219, opened 2018-04-29
Trigger loading projects on component mount if the widget doesn't hav…

ljwboschker:
To fix the double XHR requests from CA-360, I removed the fetch projects trigger on component mounting. As a result, projects were not loaded anymore when there is no controller. Fixed it by only triggering project loading if the widget has no controller.


ljwboschker:
Moved to a new branch, together with CA-389. Will create a new pull-request for this.

PR #221, opened 2018-04-29
Merge Ca 289 task into Friesland

dominicusplatus:


dominicusplatus:
friesland workflow is different

PR #232, opened 2018-05-07
Ca 382 festival

ljwboschker:
Added the Festival page, homepage banner and footer link to the site.

Not 100% sure about the FestivalService, which basically contains only one getter to determine if the Festival links should be displayed. It works, but do you have any other suggestions?


CollAction/Services/FestivalService.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using Microsoft.Extensions.Options;
+
+namespace CollAction.Services
+{
+    public class FestivalService : IFestivalService
+    {
+        private readonly FestivalServiceOptions _festivalOptions;
+        private readonly DateTime DefaultEndDate = new DateTime(2018, 5, 27, 23, 59, 59);
```
logophobia: The default enddate seems unnecessary.

I'd say if there is no date date, don't show the festival stuff. No defaults necessary

CollAction/Services/FestivalService.cs
```diff
@@ -0,0 +1,18 @@
...
+
+        public FestivalService(IOptions<FestivalServiceOptions> festivalOptions)
+        {
+            _festivalOptions = festivalOptions.Value;
+        }
+
+        public bool CtasVisible => DateTime.Now < (_festivalOptions.FestivalEndDate ?? DefaultEndDate);
```
logophobia: Probably best if you write out Ctas in full. I'm not sure what it stands for.

logophobia:
The service is fine like this. It's a bit small, but works well enough.

PR #233, opened 2018-05-13
CA-391

logophobia:
Also did a small cleanup of the project controller, lots of commented out code and redundant comments


ljwboschker:
The whole SetTags method in Project.cs seems a bit over-complicated... entity framework should be able to do these changes without need to the ApplicationDbContext here.

PR #235, opened 2018-05-13
CA-303

logophobia:
Another attempt. This time, we'll accept requests from remote proxies, something we didn't test for.


CollAction/Startup.cs
```diff
@@ -127,9 +128,74 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
...
+                        cspBuilder.AddImgSrc().Self() // Only allow self-hosted images, or google analytics (for tracking images)
+                                              .Sources.AddRange(new[]
+                                                                {
+                                                                    "https://www.google-analytics.com"
+                                                                });
+                        cspBuilder.AddStyleSrc().Self() // Only allow style/css from these sources (note: css injection can actually be dangerous)
+                                                .UnsafeInline() // Unfortunately this is necessary, the backend passess some things that are directly passed into the css, especially on the project page. TODO: We should try to get rid of this.
+                                                .Sources.AddRange(new[]
+                                                                  {
+                                                                      "https://maxcdn.bootstrapcdn.com/",
```
dominicusplatus: Could think of some way to dynamically retrieve style/font sources via provider, at this time we must ensure to remember to add to here if some new source is used in frontend.

logophobia: Problem is that currently the fonts are retrieved through some 3rd party javascript, which then loads things from another domain. Don't see an easy feasible way to do it. If you have a good refactoring, feel free to add.

PR #236, opened 2018-05-13
CA-303 Friesland

logophobia:
Another attempt. This time, we'll accept requests from remote proxies, something we didn't test for.


CollAction/Frontend/app/project/PieChart.tsx
```diff
@@ -0,0 +1,53 @@
+import * as React from "react";
```
logophobia: Rewrote the pie chart here so that it no longer contains inline javascript, and is a proper react component.

PR #239, opened 2018-05-27
EDIT: CA-289 project names correct routing with js uricomponent encode

dominicusplatus:
CA289


CollAction/Controllers/ProjectsController.cs
```diff
@@ -66,14 +67,24 @@ public ViewResult StartInfo()
...
+            if(String.IsNullOrEmpty(validUrlForProjectName))
+            {
+                return NotFound();
+            }
+            return LocalRedirect("~/projects/"+validUrlForProjectName+"/details");
+        }
+        
+        public async Task<IActionResult> Details(string name)
+        {
+            List<DisplayProjectViewModel> items = await DisplayProjectViewModel.GetViewModelsWhere(_context, p => p.Name == WebUtility.UrlDecode(name) && p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted);
```
logophobia: Things should be decoded automatically. I don't think WebUtility.UrlDecode is necessary here (maybe test that, not 100% sure).

CollAction/Controllers/ProjectsController.cs
```diff
@@ -66,14 +67,24 @@ public ViewResult StartInfo()
...
-            if (id == null)
+            var project =  await _service.GetProjectById(id); 
+            if (project == null)
             {
                 return NotFound();
             }
-
-            List<DisplayProjectViewModel> items = await DisplayProjectViewModel.GetViewModelsWhere(_context, p => p.Id == id && p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted);
+            string validUrlForProjectName = WebUtility.UrlEncode(project.Name);
+            if(String.IsNullOrEmpty(validUrlForProjectName))
```
logophobia: Don't think this check is necessary. A project must have a name. You already check if the project was found before this.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -186,12 +197,23 @@ public ViewResult StartInfo()
...
+            }
+            return LocalRedirect("~/Projects/Create/"+validUrlForProjectName+"/thankyou");
         }
 
+        [Authorize]
+        public IActionResult ThankYouCreate(string name)
+        {
+            return View(new ThankYouCreateProjectViewModel
+            {
+                Name = WebUtility.UrlDecode(name)
```
logophobia: The project name should already be decoded.

dominicusplatus: Apparently due to some reason this is not always the case ( checked it ), so it's better to decode it always. Could have something to do with redirect.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -66,14 +67,24 @@ public ViewResult StartInfo()
...
                 return NotFound();
             }
-
-            List<DisplayProjectViewModel> items = await DisplayProjectViewModel.GetViewModelsWhere(_context, p => p.Id == id && p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted);
+            string validUrlForProjectName = WebUtility.UrlEncode(project.Name);
+            if(String.IsNullOrEmpty(validUrlForProjectName))
+            {
+                return NotFound();
+            }
+            return LocalRedirect("~/projects/"+validUrlForProjectName+"/details");
```
logophobia: Using formatting might be a little more idiomatic, as in, LocalRedirect($"/Projects/{validUrlForProjectName}/Details").

CollAction/Controllers/ProjectsController.cs
```diff
@@ -66,14 +67,24 @@ public ViewResult StartInfo()
...
         {
-            if (id == null)
+            var project =  await _service.GetProjectById(id); 
+            if (project == null)
             {
                 return NotFound();
             }
-
-            List<DisplayProjectViewModel> items = await DisplayProjectViewModel.GetViewModelsWhere(_context, p => p.Id == id && p.Status != ProjectStatus.Hidden && p.Status != ProjectStatus.Deleted);
+            string validUrlForProjectName = WebUtility.UrlEncode(project.Name);
```
logophobia: Naming is a bit werid, it's not a url, it's an encoded project name.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -372,14 +394,9 @@ public ViewResult StartInfo()
...
-        public async Task<IActionResult> Commit(int? id)
+        public async Task<IActionResult> Commit(string name)
         {
-            if (id == null)
-            {
-                return NotFound();
-            }
-
-            var project =  await _service.GetProjectById(id); 
+            var project =  await _service.GetProjectByName(WebUtility.UrlDecode(name));
```
logophobia: Again, should already be decoded.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -428,6 +440,24 @@ public ViewResult StartInfo()
...
 
+        [Authorize]
+        [HttpGet]
+        public IActionResult ThankYouCommit(string name)
+        {
+            if(name!=null)
+            {
+                CommitProjectViewModel model = new CommitProjectViewModel()
+                {
+                    ProjectName = WebUtility.UrlDecode(name)
```
logophobia: Should already be decoded.

CollAction/Controllers/ProjectsController.cs
dominicusplatus: Apparently due to some reason this is not always the case ( checked it ), so it's better to decode it always. Could have something to do with redirect.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -428,6 +440,24 @@ public ViewResult StartInfo()
             }
         }
 
+        [Authorize]
+        [HttpGet]
+        public IActionResult ThankYouCommit(string name)
+        {
+            if(name!=null)
```
logophobia: Use the required attribute if you want to exclude null values.

CollAction/Controllers/ProjectsController.cs
logophobia: Do you know which circumstances this happens? Because double decoding might give rise to other issues.

Suppose a project name has something in it that happens to be a valid url encoding, then double decoding it will mean we won't find the project because the name'll be wrong.

If it's that unreliable, maybe we should go with the id + name idea. That way we don't need to mess with the name that much.

logophobia:
Merged some things that interact with the share buttons, and thus the urls. Make sure to be careful if/when you merge.

PR #242, opened 2018-05-31
CA-399 friesland

logophobia:


logophobia:
We should translate the policy at some point, but I'm not really comfortable translating it myself, legal stuff can be tricky, you can't afford errors there.

Given that gdpr is already here, we can't wait too long either.

ronvdakker:
I've asked the legal team for a Dutch version (of the version that will be final - we first need to include the previous feedback) - will let you know once they get back!

Re: styling @ljwboschker : you mean for the privacy policy itself? tnx

PR #244, opened 2018-06-01
Build docker containers for each pull request and push to ECR

ljwboschker:
Hi Tim, no need to approve right now; I need a pull request to see if I can get Travis-CI to push docker containers to ECR for each pull request. This is still work-in-progress.


logophobia:
Ah, that would be really nice.

ljwboschker:
@logophobia I got it working. I also created some documentation on [https://collaction.atlassian.net/wiki/spaces/CATS/pages/523337732/Deploy+to+test.collaction.org]. 

Can you take a look at it, see if I missed something, or if things can be improved?

Only downside I haven't been able to solve is that it uses the PR-number as container tag. I would have preferred the Jira issue number, but that one isn't available in Travis. Any ideas on that?

logophobia:
Might also be a good idea to host the database on RDS, so that the test data is persistent. 

Now, you need to keep starting projects again and again, because the database is dropped when you start a new task on the project.

ljwboschker:
The script now checks the branch name for a "CA-n" number. If it contains one, it will build and push a container to ECR using the CA-number as image tag. You can also use ./deploy.sh/ps1 script to deploy the image using it's CA-number to the test-cluster.

That way the containers reference the issue number instead of the PR-number.

PR #245, opened 2018-06-03
CA-289 friesland

dominicusplatus:


CollAction/Controllers/ProjectsController.cs
```diff
@@ -311,6 +317,25 @@ public IActionResult Embed()
             }
         }
 
+        [Authorize]
+        [HttpGet]
+        public IActionResult ThankYouCommit(string name)
+        {
+            if(name!=null)
```
logophobia: Formatting. Make sure you "space" things out.

logophobia:
Merged some things that interact with the share buttons, and thus the urls. Make sure to be careful if/when you merge.

PR #246, opened 2018-06-16
CA-401 Cookie Policy

ljwboschker:
Show cookie message banner on every page, until the 'Accept' button has been clicked. Acceptance state is stored in a cookie on the domain the site is running on.

### DIFF
--- a/CollAction/Frontend/app/global/CookieMessage.tsx
+++ b/CollAction/Frontend/app/global/CookieMessage.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import renderComponentIf from "./renderComponentIf";
 
 interface ICookieMessageProps {
-    showBanner: boolean;
+    canTrack: boolean;
     cookieString: string;
 }
 
@@ -18,16 +18,28 @@ export default class CookieMessage extends React.Component<ICookieMessageProps, 
 
     componentDidMount() {
         this.setState({
-            showBanner: this.props.showBanner
+            showBanner: !this.props.canTrack
         });
+
+        if (this.props.canTrack === true) {
+            this.giveCookieConsent();
+        }
     }
 
     accept() {
         document.cookie = this.props.cookieString;
+        this.giveCookieConsent();
         this.setState({
             showBanner: false
         });
     };
+
+    giveCookieConsent() {
+        let dataLayer = window["dataLayer"];
+        if (dataLayer) {
+            dataLayer.push({"event": "cookie_consent_given"});
+        }
+    }
 
     renderMessage() {
         return (
@@ -58,7 +70,7 @@ export default class CookieMessage extends React.Component<ICookieMessageProps, 
 
 renderComponentIf(
     <CookieMessage
-        showBanner={ document.getElementById("cookie-message") && document.getElementById("cookie-message").dataset.showBanner === "True" }
+        canTrack={ document.getElementById("cookie-message") && document.getElementById("cookie-message").dataset.canTrack === "True" }
         cookieString= { document.getElementById("cookie-message") && document.getElementById("cookie-message").dataset.cookieString }
     />,
     document.getElementById("cookie-message")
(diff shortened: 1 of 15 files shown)


logophobia:
It might be a good idea to use the gdpr middleware to block cookies until we get consent:

https://docs.microsoft.com/en-us/aspnet/core/security/gdpr?view=aspnetcore-2.1

If you upgrade to asp 2.1, then you also solved CA-416.

Also, we should pair this with google tag manager modifications, to block google analytics until we get consent:

https://www.metric.ai/blog/gdpr-cookies

And we should probably block:
* Social media
* Share / Like
* Youtube

Until we get consent.

ljwboschker:
@logophobia can you take a look at the latest changes?

Following is _not_ loaded until the cookie policy is accepted:
* Google Tag Manager
* Share buttons (homepage, 2x on project page and 1x on thank you page)
* Social media login options

Did I miss something?

logophobia:
Also, if you set all cookies to needs consent, then I don't think they'll be able to login at all. So maybe block all of the login?

CollAction/Views/Shared/_Layout.cshtml
```diff
@@ -1,11 +1,20 @@
-﻿<!DOCTYPE html>
+﻿@using Microsoft.AspNetCore.Http.Features
+
+@{
+    var consentFeature = Context.Features.Get<ITrackingConsentFeature>();
+    var canTrack = consentFeature?.CanTrack ?? false;
```
logophobia: I don't think you need the null coalesce here, if the consent feature isn't set, it should be an error right?

logophobia:
Maybe in version 2 we could have options, like don't allow anonymous tracking, but do allow stuff for the essentials (like login).

CollAction/Frontend/app/global/CookieMessage.tsx
```diff
@@ -0,0 +1,63 @@
...
+    };
+
+    renderMessage() {
+        return (
+            <div>
+                <div id="cookie-message-banner">
+                    <div className="container">
+                        <div className="row">
+                            <div className="col-xs-12">
+                                <p>Thank you for visiting our website! Please note that our website uses cookies to analyze and improve the performance of our website and to make social media integration possible. For more information on the use of cookies and privacy related matters, please see our <a href="home/privacy">Privacy and Cookies Policy</a>. By clicking "I accept" or by using this site, you consent to the use of cookies.</p>
```
logophobia: I think this is still missing a few things. I thought this was a good summary: https://www.cookiebot.com/en/gdpr-cookies/

* User should be able to reject the non-essential cookies (for example: tracking, youtube)
* I think this should be a bit clearer, maybe a bullet point list with a link to the full privacy policy?

CollAction/Views/Account/Login.cshtml
```diff
@@ -10,6 +11,8 @@
 
 @{
     ViewData["Title"] = Localizer["Log in"];
+    var consentFeature = Context.Features.Get<ITrackingConsentFeature>();
```
logophobia: Feels like it should be part of the viewmodel.

ljwboschker: I'm not sure.. I got this from the Microsoft documentation itself. 

Also, there are 3 parts (index, _layout and _cookieConsentPartial) where there is no view model. I could introduce one for the index page, but not for _layout and _cookieConsentPartial. So we'll end up with two ways where this is set.

CollAction/Frontend/app/global/CookieMessage.tsx
ljwboschker: @ronvdakker can you comment on this? See if we need different copy here.

Do we want to give users an option to either opt-in or opt-out of tracking cookies. Perhaps you can create a new story for that?

ronvdakker: Yes, we can create a separate story for the opt in and opt out of essentials. 

Lets go ahead with this one for now though, and improve upon it later.

ljwboschker:
I checked the site without consent (including login and admin pages) and it seems to still work. 

Non-essential cookies are TempData (only used on social login) and session state (which we don't use). So it should work. Nevertheless, some extra testing of the site with the banner visible is in order.

CollAction/Views/Account/Login.cshtml
logophobia: Alright, didn't see that. It's fine as it is then.

PR #251, opened 2018-06-22
CA-440 upgrade dotnet 2.1 and nuget packages

ljwboschker:


ljwboschker:
The docker-compose.yml itself doesn't have a reference to images, those are in ./CollAction/Dockerfile. I upgraded that one.

Visual Studio 2017 does advise you to upgrade to the latest version when using dotnet 2.1. I'll post a message in the general slack channel about that.

logophobia:
@ljwboschker Luuk, I do think this needs to be tested before you merge! Like everything! I've upgraded 2.1 before, the gdpr stuff it might block cookies of some features by default, so it's a good idea to do an acceptance test!

ljwboschker:
Yes, my bad. Thought of that myself. Merge is reverted, test-version is now running.

PR #257, opened 2018-07-07
Add application insights CA-454

logophobia:


ljwboschker:
Probably better to add it only for production-environments, to prevent dev environments from sending data to this resource.

Also, this needs an instrumentation key. We might want have two; one for test-environment(s) and one for production.

CollAction/Startup.cs
```diff
@@ -108,6 +108,7 @@ public void ConfigureServices(IServiceCollection services)
                 options.MailChimpKey = Configuration["MailChimpKey"];
                 options.MailChimpNewsletterListId = Configuration["MailChimpNewsletterListId"];
             });
+            services.AddApplicationInsightsTelemetry(Configuration);
```
ljwboschker: Probably better to add it only for production-environments, to prevent dev environments from sending data to this resource.

PR #258, opened 2018-07-07
Add application insights CA-453

logophobia:


CollAction/Startup.cs
```diff
@@ -112,6 +112,7 @@ public void ConfigureServices(IServiceCollection services)
                 // Note: use yyyy-MM-dd HH:mm:ss notation. Don't specify to use the default value (2018-05-27 23:59:59)
                 options.FestivalEndDate = Configuration.GetValue<DateTime?>("FestivalEndDate");
             });
+            services.AddApplicationInsightsTelemetry(Configuration);
```
ljwboschker: Probably better to add it only for production-environments, to prevent dev environments from sending data to this resource.

PR #263, opened 2018-07-15
CA-433 - allow project owners to send e-mails

logophobia:
This turned out a little bigger than I hoped.

* Because you're sending customized e-mails, you need a lot of external API calls for one request. We needed some form of background-queueing, so I integrated hangfire. Persistent background queueing system, it comes with monitoring dashboard. Also has build in retrying, this makes our SES API calls resilient against downtime. Might be a bit heavyweight.
* Since we want most of our external API calls to be resilient, I also wrapped mailchimp in hangfire. If we unsubscribe, we really don't want to fail there.
* We've also got unsubscribe functionality, with 128-bit tokens stored in the database

Also fixed some small stuff that shouldn't change things externally, I can move it to a separate PR if people think it necessary:
* Had to do a lot of stuff in the projects controller, I noticed that:
  - There was a lot of unused stuff there (viewmodels), removed that
  - Some weird naming conventions which made me misunderstand the purpose of some functionality, did a little bit of renaming to clarify things, we'll have to be careful testing (I did retest everything)
  - A duplicate expensive query in the find functionality, which is our "hot path". Very expensive, removed the duplicate.
  - Misconfigured project creation validation that wouldn't always show all errors
* Made the admin role a constant
* SetPassword was unused and never linked anywhere. Untested hidden stuff might become security issue when unmaintained (not atm though)


CollAction/Views/Projects/SendProjectEmail.cshtml
```diff
@@ -0,0 +1,37 @@
...
+@model SendProjectEmail
+@{
+    ViewData["Title"] = $"Send project e-mail for '{Model.Project.Name}'";
+}
+
+<div id="project-email-form">
+    <div class="container">
+        <h2>@ViewData["Title"]</h2>
+        <p>
+            You have already send @Model.Project.NumberProjectEmailsSend e-mails, and can still send @Model.EmailsAllowedToSend emails up until @Model.SendEmailsUntil. No HTML is allowed, {firstname} and {lastname} will be substituted with the user's first and last name.
```
dominicusplatus: already sent ?

CollAction/Models/ProjectViewModels/FindProjectsViewModel.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        public int Target { get; set; }
+
+        public int Participants { get; set; }
+
+        public int ProgressPercent
+        {
+            get
+            {
+                return Participants * 100 / Target;
```
dominicusplatus: Is target default initialized to >0 somewhere so that we won't get exception for division by 0 ?

dominicusplatus: Ok project cannot be created with 0 target I think by validation but could be possible to update it by admin to 0 ?

CollAction/Services/ProjectEmailOptions.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace CollAction.Services
+{
+    public class ProjectEmailOptions
+    {
+        public int MaxNumberProjectEmails { get; set; }
+        public TimeSpan MaxTimeProjectEmails { get; set; }
```
ljwboschker: After looking it up, I think this means how long the owner can send e-mails after the project ended. Maybe a more descriptive name; TimeAllowedAfterProjectEnds?

CollAction/Services/ProjectEmailOptions.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace CollAction.Services
+{
+    public class ProjectEmailOptions
+    {
+        public int MaxNumberProjectEmails { get; set; }
```
ljwboschker: I think we need a default for this (probably equal to the maximum number of participants allowed on a project), or don't allow the application to start if it's not set. Now it defaults to 0, which means this function won't work. Something that could easily go wrong some time...

CollAction/Services/ProjectService.cs
```diff
@@ -64,26 +72,40 @@ public ProjectService(ApplicationDbContext context, IStringLocalizer<IProjectSer
...
+                        Target = project.Target,
+                        Participants = project.Participants.Sum(participant => participant.User.RepresentsNumberParticipants) + project.AnonymousUserParticipants,
+                        Remaining = project.RemainingTime,
+                        DescriptiveImagePath = project.DescriptiveImage.Filepath,
+                        DescriptiveImageDescription = project.DescriptiveImage.Description,
+                        Status = project.Status,
+                        Start = project.Start,
+                        End = project.End
+                    })
+                .ToListAsync();
```
ljwboschker: This probably generates a very complex query, in a method that's called a lot. You might want to profile it, see if the database can take it.

In my experience, a limited set if simple queries can perform a lot better then a very complex one, even if it does require more roundtrips.

logophobia: True, we should do some optimization.

This however isn't a new query, it's the exact same query that was already there, things just moved around a little.

Maybe we should create a seperate ticket for that, I didn't change anyhere except the naming of the viewmodels.

CollAction/Services/ProjectEmailOptions.cs
logophobia: Changed it

CollAction/Services/ProjectEmailOptions.cs
logophobia: Added some defaults

CollAction/Models/ProjectViewModels/FindProjectsViewModel.cs
logophobia: I checked, the admin can't update it to 0 as well. This is already how it works, just moved it around little.

CollAction/Views/Projects/SendProjectEmail.cshtml
logophobia: Corrected, thank you

PR #270, opened 2018-07-19
CA-289 | 	CA-388 Friesland

dominicusplatus:


CollAction/Controllers/ProjectsController.cs
```diff
@@ -200,12 +208,20 @@ public IActionResult Embed()
...
+            return LocalRedirect($"~/Projects/Create/{validUriPartForProjectName}/thankyou");
         }
 
+        [Authorize]
+        public IActionResult ThankYouCreate(string name)
+        {
+            return View(new ThankYouCreateProjectViewModel
+            {
+                Name = name
+             });
```
logophobia: Watch the formatting

CollAction/Controllers/ProjectsController.cs
```diff
@@ -311,6 +317,25 @@ public IActionResult Embed()
             }
         }
 
+        [Authorize]
+        [HttpGet]
+        public IActionResult ThankYouCommit(string name)
+        {
+            if(name != null)
```
logophobia: Formatting again, should be "if ("

Are you using visual studio? These things should auto-correct.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -311,6 +317,25 @@ public IActionResult Embed()
...
+            {
+                CommitProjectViewModel model = new CommitProjectViewModel()
+                {
+                    ProjectName = name
+                };
+                return View("ThankYouCommit", model);
```
logophobia: nameof(ThankYouCommit) would be robust again action renames.

CollAction/Views/Projects/ThankYouCommit.cshtml
```diff
@@ -13,10 +13,9 @@
     <div class="row">
         <div id="thank-you-commit-box" class="col-xs-12 col-md-6 col-md-offset-3">
             <div id="close-button-container">
-                <a id="close-button" asp-action="Details" asp-route-id="@Model.ProjectId">
-                    <object id="close-button-img" type="image/svg+xml" data="/images/close-button-friesland-blue.svg">
-                        @*Close*@
-                        Sluiten
+                <a id="close-button" asp-action="Details" asp-route-id="@Model.ProjectName">
+                    <object id="close-button-img" type="image/svg+xml" data="/images/close-button.svg">
+                        Close
```
logophobia: You should keep "Sluiten" if it's going to be merged to Friesland. And keep close-button-friesland-blue.svg.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -60,14 +61,21 @@ public ViewResult StartInfo()
             return View(model);
         }
 
-        public async Task<IActionResult> Details(int? id)
+        // GET: Projects/Details/5
+        public async Task<IActionResult> DetailsById(int? id)
```
ljwboschker: Any particular reason why it's a nullable int?

CollAction/Controllers/ProjectsController.cs
dominicusplatus: VS code does not show this formatting errors by default

CollAction/Controllers/ProjectsController.cs
dominicusplatus: Not really, only that GetProjectById takes nullable int so it's type compatible already.

CollAction/Views/Projects/ThankYouCommit.cshtml
dominicusplatus: Fixed in 0544b69

CollAction/Controllers/ProjectsController.cs
dominicusplatus: Fixed in 0544b69

CollAction/Controllers/ProjectsController.cs
dominicusplatus: Fixed in 0544b69

CollAction/Controllers/ProjectsController.cs
dominicusplatus: Fixed in 0544b69

CollAction/Controllers/ProjectsController.cs
logophobia: Null is probably not a sensible parameter, since it'll only yield a 404. It being nullable is probably a small pre-existing bug. Good idea to fix as well.

PR #271, opened 2018-07-29
CA-308

dominicusplatus:


CollAction/Startup.cs
```diff
@@ -119,6 +120,25 @@ public void ConfigureServices(IServiceCollection services)
                 options.Password.RequireNonAlphanumeric = false;
                 options.Password.RequiredLength = 8;
             });
+            
+            services.Configure<GzipCompressionProviderOptions>(options => options.Level = System.IO.Compression.CompressionLevel.Optimal);
```
logophobia: That's a very high level. Remember, the compression level is a trade-off between cpu used and reduction in network traffic. Optimal means the server load might go up quite a bit, especially under load. 

Best play around a bit with that option, see what level gets us the most gain for the least load increase. Plot it?

dominicusplatus: Ok I can check all compression levels and see what results it gives and if there is not much difference we could use lowest compression, but I don't know how to reliably benchmark it's CPU load that might be tricky, I would have to inject somehow PerformanceCounter into compression middleware to really is what it takes, which high speed CPUs and couple KBs it might be less the miliseconds time.

logophobia: If it's too difficult then it's not necessary to measure CPU.

Just see what level gives us the highest gains. For example, if there's hardly a difference between Optimal and Default wrt to time saved, then use the default level. My guess is that you'll hardly see a difference once you go above a certain compression level.

We can do CPU measurements if the previous test is not sufficient.

logophobia: You can also do 

    services.Configure<GzipCompressionProviderOptions>(Configuration);

And make the level configurable.

PR #272, opened 2018-07-29
CA-430

dominicusplatus:
…24h but <48h describe 1 day and hours


CollAction/Models/ProjectViewModels/DisplayProjectViewModel.cs
```diff
@@ -86,6 +86,8 @@ public string RemainingTimeUserFriendly
                     return $"{remaining.Weeks()} weeks";
                 else if (remaining.Days > 1)
                     return $"{remaining.Days} days";
+                else if (remaining.TotalHours >= 24)
```
logophobia: Why not, instead of:

    else if (remaining.TotalHours >= 24)
        return $"{(int)remaining.TotalHours} hours";
    else if (remaining.Hours > 1)
        return $"{remaining.Hours} hours";

do this:
    else if (remaining.Hours > 1)
        return $"{remaining.TotalHours} hours";

The two if-statements are a bit redundant this way.

dominicusplatus: ok, but cast is necessary, totalhours is actually double ( https://docs.microsoft.com/pl-pl/dotnet/api/system.timespan.totalhours?view=netframework-4.7.2#System_TimeSpan_TotalHours ) and it displays 21.32421.

dominicusplatus: fixed

PR #274, opened 2018-07-29
CA-481

dominicusplatus:


CollAction/Frontend/app/global/MobileNavDraw.tsx
```diff
@@ -36,6 +36,10 @@ export default class MobileNavDraw extends React.Component<IMobileNavDrawProps,
             <li><a href="/find">Find Project</a></li>
             <li><a href="/start">Start Project</a></li>
             <li><a href="/about">About</a></li>
+            {this.props.loginText == "Logout" ?
```
logophobia: This whole thing could be solved a bit more neatly. There are two if statements (checking for "Logout"), and they're beneath each other.

First, make these one proper if statement, not these weird ternary statements: https://reactjs.org/docs/conditional-rendering.html

It's a pre-existing problem with this code, but loginText should be a boolean, since it's treated that way. Maybe call it "isLoggedIn"?

CollAction/Frontend/app/global/MobileNavDraw.tsx
```diff
@@ -36,10 +53,8 @@ export default class MobileNavDraw extends React.Component<IMobileNavDrawProps,
             <li><a href="/find">Find Project</a></li>
             <li><a href="/start">Start Project</a></li>
             <li><a href="/about">About</a></li>
-            {this.props.loginText == "Logout" ?
-                (<li><button type="button" onClick={() => document.getElementById("logOutBtn").click()}>Logout</button></li>) :
-                (<li><a href="/account/login">Login</a></li>)
-            }
+            <AccountManageNav isLoggedIn={this.props.isLoggedIn} />
```
logophobia: This works, but there's a much simpler solution, inside render, add this variable:

```
let accountHtml;

if (this.props.isLoggedIn) {
    accountHtml = <li><a href="/manage">Account</a></li><li><button type="button" onClick={() => document.getElementById("logOutBtn").click()}>Logout</button></li>;
} else {
    accountHtml = <li><a href="/account/login">Login</a></li>;
}
```

and then

```
return(
      <div id="draw-menu">
        <div id="draw-menu-close-button" onClick={() => self.close()}>
          <i className="fa fa-times" aria-hidden="true"></i>
        </div>
        <ul>
            {accountHtml}
        </ul>
      </div>
);
```

I'm fine with the component based solution, this is a lot simpler though.

CollAction/Frontend/app/global/MobileNavDraw.tsx
```diff
@@ -3,12 +3,29 @@ import renderComponentIf from "./renderComponentIf";
...
+function AccountManageNav(props) {
+  if (!!props.isLoggedIn) {
+    return <li><a href="/manage">Account</a></li>;
+  }else {
+    return null;
+  }
+ }
+
+ function LoginStateNav(props) {
+  if (!!props.isLoggedIn) {
```
logophobia: Bit unrelated, but you don't need to use the double negatives if you actually declare isLoggedIn as boolean. This is typescript, so best we use typing as much as possible to make the code more robust.

PR #277, opened 2018-08-03
CA-383 start project form

ljwboschker:
The new project startup-page.


ljwboschker:
A few notes:

- I had to make some changes to the CSP rules to get image preview and client side validation in combination with webpack/typescript working
-- Should be safe, because there is XSS protection on all input fields
- the client side validation scripts (jquery and microsoft scripts) are loaded from a CDN, instead of from the lib directory
-- @dominicusplatus I think you were looking at the lib scripts as well. We'll hvae to sync on that when merging

CollAction/Frontend/app/global/RichTextRequiredValidation.ts
```diff
@@ -0,0 +1,25 @@
+if (typeof $ !== "undefined" && typeof $.validator !== "undefined" && typeof $.validator.unobtrusive !== "undefined") {
```
logophobia: Wouldn't a better place for this be inside the rich-text editor component, in the render function?

CollAction/Frontend/app/global/WithinMonthsValidation.ts
```diff
@@ -0,0 +1,52 @@
+import * as moment from "moment";
+
+if (typeof $ !== "undefined" && typeof $.validator !== "undefined" && typeof $.validator.unobtrusive !== "undefined") {
+
+    $.validator.addMethod("withinmonthsaftertoday",
```
logophobia: This would a bit more readable if the string used camel-casing or something similar.

CollAction/Frontend/app/project/UploadDescriptiveImage.tsx
```diff
@@ -0,0 +1,94 @@
...
+        this.getFileInputElement = this.getFileInputElement.bind(this);
+        this.removeBanner = this.removeBanner.bind(this);
+    }
+
+    onDrop(accepted: File[], rejected: File[], event: any) {
+        super.loadImage(accepted, rejected, event);
+    }
+
+    previewImage(type: string, image: any) {
+        let background = document.getElementById("preview") as HTMLImageElement;
```
logophobia: This way of doing things feels like going against the point of react a bit. If you have type and image as part of your state, and then putting this in the render function, then you don't even need fildling with document.getElement, it's just a re-render (which should be fast with the virtual dom).

If image and type are part of the state

CollAction/Frontend/app/project/UploadDescriptiveImage.tsx
```diff
@@ -0,0 +1,94 @@
...
+        super.rejectImage();
+    }
+
+    getFileInputElement(): HTMLInputElement {
+        return document.getElementsByName("DescriptiveImageUpload")[0] as HTMLInputElement;
+    }
+
+    removeBanner() {
+        super.resetImage();
+        let background = document.getElementById("preview") as HTMLImageElement;
```
logophobia: Same again, should just be a re-render.

CollAction/Startup.cs
```diff
@@ -201,7 +202,8 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
                                                    .Sources.AddRange(Configuration["CspMediaSrc"]?.Split(";") ?? new string[0]); // Only allow self-hosted videos, or configured sources
                            cspBuilder.AddFrameAncestors().None(); // No framing allowed here (put us inside a frame tag)
                            cspBuilder.AddScriptSrc() // Only allow scripts from our own site, the aspnetcdn site, app insights and google analytics
-                                     .Self()
+                                     .Self()                                    
+                                     .UnsafeEval()  // Needed for client side validation scripts
```
logophobia: Is there no different way to solve this? Not having UnsafeEval does protect us against a whole slew of vulnerabilities that we could get if the rich text editor is in some way "leaky" (XSS), it's important for the whole defense-in-depth thing.

We could whitelist the validations if you have a list for example.

CollAction/ValidationAttributes/RichTextRequiredAttribute.cs
```diff
@@ -0,0 +1,16 @@
+using System.ComponentModel.DataAnnotations;
```
logophobia: If this is part of the rich text validation, shouldn't this be added to the SecureRichTextAttribute?

CollAction/Frontend/app/project/UploadBanner.tsx
```diff
@@ -0,0 +1,77 @@
...
+        this.removeBanner = this.removeBanner.bind(this);
+    }
+
+    onDrop(accepted: File[], rejected: File[], event: any) {
+        super.loadImage(accepted, rejected, event);
+    }
+
+    previewImage(type: string, image: any) {
+        let background = document.getElementById("project-background");
+        background.style.backgroundImage = "url(data:" + type + ";" + image + ")";
```
logophobia: Same again, should just be a re-render.

CollAction/Frontend/app/project/UploadBanner.tsx
```diff
@@ -0,0 +1,77 @@
...
+        super.rejectImage();
+    }
+
+    getFileInputElement(): HTMLInputElement {
+        return document.getElementsByName("BannerImageUpload")[0] as HTMLInputElement;
+    }
+
+    removeBanner() {
+        super.resetImage();
+        let background = document.getElementById("project-background");
```
logophobia: Same again, should just be a re-render.

CollAction/package.json
```diff
@@ -3,33 +3,40 @@
...
-    "react-tap-event-plugin": "^2.0.1",
-    "sass-loader": "^4.1.1",
-    "scss": "^0.2.4",
-    "style-loader": "^0.13.1",
-    "ts-loader": "^1.3.3",
-    "tslint": "^4.4.2",
-    "typescript": "^2.1.5",
-    "webpack": "^1.14.0"
+    "@types/jquery": "^3.3.5",
+    "@types/jquery-validation-unobtrusive": "^3.2.32",
```
logophobia: If we're loading unobtrusive validation through npm, do we still the cdn links?

CollAction/Views/Shared/_Layout.cshtml
```diff
@@ -58,7 +58,7 @@
                             <li><a asp-area="" asp-controller="Projects" asp-action="Find">Find Project</a></li>
                             <li><a asp-area="" asp-controller="Projects" asp-action="StartInfo">Start Project</a></li>
                             <li><a asp-area="" asp-controller="Home" asp-action="About">About</a></li>
-                            @if (User.IsInRole(Constants.AdminRole))
+                            @if (User.IsInRole("Admin"))
```
logophobia: Why? The constants ensure the admin role is "typo-checked".

CollAction/Frontend/app/project/UploadDescriptiveImage.tsx
ljwboschker: I see what you mean; not sure if it's a good idea since images can take up a lot of memory. I'll profile it to see what making it reactive means for browser memory usage.

CollAction/Startup.cs
ljwboschker: I'll take a look. I'm not really happy with how this works either; the validators aren't really part of the typescript react code but end up in the bundle non-the-less. I'll see if I can find another solution. Probably vanilla javascript.

CollAction/ValidationAttributes/RichTextRequiredAttribute.cs
ljwboschker: Not all rich text components are mandatory :-) So I think it's nice to keep the validation attributes separate.

CollAction/Views/Shared/_Layout.cshtml
ljwboschker: A that slipped in.. it's wouldn't compile on my machine. No idea why; probably a version mixup. I'll put some more time in that.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -67,8 +67,12 @@ public async Task<IActionResult> Details(int? id)
         [Authorize]
         public async Task<IActionResult> Create()
         {
+            var user = await  _userManager.GetUserAsync(User);
```
logophobia: 2 spaces between await and _userManager

CollAction/ValidationAttributes/RichTextRequiredAttribute.cs
logophobia: That probably works then. What is this doing differently from the normal [Required] attribute, client side validation?

ljwboschker: This one forces JQuery Validation to use the 'richtextrequired' client side validator instead if it's own default validator. The difference with the client side validator is that checks for actual content. 

For example: when pressing enter in the rich text editor, you'll get the following content: `<p></p>`. Which is valid content according to the default validator.

The rich text validator strips away the HTML attributes (thanks to `$(value).text().trim()`) and then checks if there is content.

CollAction/Frontend/app/global/RichTextRequiredValidation.ts
ljwboschker: This validation function is automatically added to the <input type="hidden"> element that is bound to a ViewModel property with a [RichTextRequired] attribute. It's not part of the rich text editor, but of jquery validation.

CollAction/Frontend/app/global/WithinMonthsValidation.ts
ljwboschker: Indeed. Done!

CollAction/Controllers/ProjectsController.cs
```diff
@@ -67,8 +67,12 @@ public async Task<IActionResult> Details(int? id)
...
+
             return View(new CreateProjectViewModel
             {
+                ProjectStarterFirstName = user?.FirstName,
```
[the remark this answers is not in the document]
ljwboschker: True. Replaced the null-coalescion with a check on the user (since it can return null according to the docs, if the user has no UserIdClaimType). But that would be an invalid operation. Throwing exception instead.

CollAction/Frontend/app/project/UploadDescriptiveImage.tsx
ljwboschker: Using reactive state now. Looks a lot better.

ljwboschker: And couldn't find a lot of difference in memory usage.

CollAction/Frontend/app/project/UploadDescriptiveImage.tsx
ljwboschker: Agreed. Same for the banner.

CollAction/Frontend/app/project/UploadBanner.tsx
ljwboschker: Agreed

CollAction/Views/Shared/_Layout.cshtml
ljwboschker: Compiles now... no idea why.

CollAction/Startup.cs
ljwboschker: The UnsafeEval is needed when the Webpack bundle is built in development mode, which is what the docker image and travis do. I'll change it so it always builds in production mode. That makes the bundle a lot smaller too.

ljwboschker:
I changed the dockerfile so that the "npm install" and "npm run build" now run in a separate stage, based on a node image. It also runs in production mode now, creating a much smaller bundle.js file and fixing the need for UnsafeEval() in the content security policy.

I also optimized the dockerfile so it makes better usage of layer-caching, which makes subsequent builds a lot faster.

ljwboschker:
Build probably fails due to service disruption in the npm registry: https://status.npmjs.org/incidents/v22ffls5cd6h

CollAction/package.json
ljwboschker: Moved the validation code into the typescript bundle, so loading from CDN is no longer needed. If also solves the need for "if $ != undefined" statements in the validation scripts.

ljwboschker:
@logophobia review comments have been processed. Could you check the updates?

.travis.yml
```diff
@@ -1,12 +1,12 @@
-language: csharp
-dotnet: 2.1.301
-mono: none
-dist: trusty
+language: minimal
```
logophobia: It's fine like this, in the S3 image pull request I'll be running unit tests in travis, so I might revert this at some point, fyi.

CollAction/Controllers/ProjectsController.cs
```diff
@@ -67,8 +67,14 @@ public async Task<IActionResult> Details(int? id)
         [Authorize]
         public async Task<IActionResult> Create()
         {
+            var user = await _userManager.GetUserAsync(User);
+
+            if (user == null) throw new InvalidOperationException("User doesn't have a UserIdClaimType. Can't determine user information.");
```
logophobia: This honestly shouldn't be possible for system-created users, not that this check is wrong, but it might be a good idea to either eventually add it in more places, or not use it at all.

CollAction/Dockerfile
```diff
@@ -1,22 +1,32 @@
...
 WORKDIR /src
 COPY CollAction/CollAction.csproj CollAction/
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
-RUN apt-get update && apt-get install -y nodejs build-essential
 RUN dotnet restore CollAction/CollAction.csproj
 COPY . .
 WORKDIR /src/CollAction
-RUN dotnet build CollAction.csproj -c Release -o /app
+RUN dotnet build CollAction.csproj --no-restore -c Release -o /app
+RUN dotnet publish CollAction.csproj --no-build -c Release -o /app
```
logophobia: I don't think travis caches the docker images in any way (fresh vm each time), so why not just do this:

RUN NODE_ENV=production dotnet build CollAction.csproj -c Release -o /app

This seems like a lot of code for a small speedup that probably doesn't really matter. 

Using the alphine image is a good idea, but I'm not really sure that the rest really matter.

ljwboschker:
Some final changes; 
- use one docker image for typescript and c# builds through csproj
- bluebird in installed using package.json instead of during project build
- put the validation logic (it's pretty big) into a separate bundle and load only on pages where it's needed
- on IE 11 it now displays a thumbnail of the banner (since background-image: url(data:...) doesn't work in IE11).

PR #279, opened 2018-08-05
CA-497 Fix csp issues newsletter subscription

logophobia:
The subscribe/unsubscribe of the newsletter had issues due to the CSP policy, this is the fix.

- Merged mailchimpmanager into the service, otherwise no queueing is possible


CollAction/Controllers/ManageController.cs
```diff
@@ -104,7 +104,7 @@ public async Task<IActionResult> Index(IndexViewModel model)
 
         [HttpPost]
         [ValidateAntiForgeryToken]
-        public async Task<IActionResult> NewsletterSubscription(IndexViewModel model)
+        public async Task<IActionResult> NewsletterSubscription([Bind("NewsletterSubscription")]IndexViewModel model)
```
ljwboschker: Why is the Bind() needed here? Won't it work with a [HttpPost("NewsletterSubscription")]?

logophobia: It's precautionary. This will ensure we only set NewsletterSubscription from the parameters in model. Prevents people from loading other variables, some of which we might actually start using in some future version of the function.

PR #281, opened 2018-08-08
Ca 476 - Move images to S3

logophobia:
* Moved each service to a separate namespace
* Unit tests are now run in the travis build (!!). I fixed the currently failing tests for that.
* There's S3 Migration code that will only need to run once, and should then be removed


CollAction/Services/Image/ImageService.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        public void DeleteImage(ImageFile imageFile)
+        {
+            if (imageFile != null)
+                _jobClient.Enqueue(() => 
+                    DeleteObject(imageFile.Filepath));
+        }
+
+        public string GetUrl(ImageFile imageFile)
+            => $"https://s3.{_region}.amazonaws.com/{_bucket}/{imageFile.Filepath}";
```
dominicusplatus: Could possibly abstract this url creation to be amazon independent taking it from config parameter, or make service named "AmazonImageService" to show that it is amazon specific storage ?

PR #282, opened 2018-08-09
CA-477 - Move images to S3

logophobia:
- Moved each service to a separate namespace
- Unit tests are now run in the travis build (!!). I fixed the currently failing tests for that.
- There's S3 Migration code that will only need to run once, and should then be removed


CollAction.Tests/ImageServicesTests.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace CollAction.Tests
+{
+    [TestClass]
+    public sealed class ImageServicesTests
```
dominicusplatus: Could may be name tests to to show it's integration tests and not run it normally with other unit tests but on demand ?

CollAction/Services/Image/ImageService.cs
```diff
@@ -0,0 +1,153 @@
...
+using Amazon;
+using Hangfire;
+using Microsoft.EntityFrameworkCore;
+using System.Linq;
+using CollAction.Helpers;
+using Microsoft.AspNetCore.Hosting;
+
+namespace CollAction.Services.Image
+{
+    public class ImageService : IImageService
```
dominicusplatus: Service could have Amazon specific name ?

CollAction.Tests/ImageServicesTests.cs
logophobia: I added a test category ("Integration") and moved all the integration tests to their own namespace. All we have is integration tests though, so better start adding unit tests.

CollAction/Services/Image/ImageService.cs
logophobia: It's AmazonS3ImageService now, interface name remains the same.

CollAction/Startup.cs
```diff
@@ -332,12 +332,17 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
...
-                    context.Database.ExecuteSqlCommand("DROP SCHEMA public CASCADE; CREATE SCHEMA public;");
-                context.Database.Migrate();
-                Task.Run(() => context.Seed(Configuration, userManager, roleManager)).Wait();
+                Task.Run(async () =>
+                {
+                    if (Configuration.GetValue<bool>("ResetTestDatabase"))
+                        await context.Database.ExecuteSqlCommandAsync("DROP SCHEMA public CASCADE; CREATE SCHEMA public;");
+                    await context.Database.MigrateAsync();
+                    await context.Seed(Configuration, userManager, roleManager);
+                    await imageService.MigrationToS3(); // TODO: Remove after it's run
```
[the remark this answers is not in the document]
logophobia: Can't get all the required services in an EF migration. Besides, there is a guard against running it again, it only executes the migration for images inside the directory "usercontent". They're moved outside that directory when migrating.

CollAction/Services/Image/AmazonS3ImageService.cs
```diff
@@ -0,0 +1,153 @@
...
+                              .ToListAsync();
+
+            foreach (var image in imagesToMigrate)
+            {
+                using (FileStream imageStream = File.OpenRead(_environment.WebRootPath + image.Filepath))
+                {
+                    Image<Rgba32> memoryImage = SixLabors.ImageSharp.Image.Load(imageStream);
+                    byte[] imageBytes = ConvertImageToPng(memoryImage);
+                    image.Filepath = $"{Guid.NewGuid()}.png";
+                    _jobClient.Enqueue(() =>
```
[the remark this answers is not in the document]
logophobia: That's a good point, I'll force the upload to happen inside the migration.

PR #293, opened 2018-09-09
Added warning text for google login issue

lucgeurts:


CollAction/Views/Account/Login.cshtml
```diff
@@ -46,8 +46,16 @@
                                 </button>
                             </div>
                         }
-                    </form>
+                    </form>                
+                    
+                    <p>
```
ljwboschker: We should only show this on the mobile viewports; on desktop it will work just file. You can use the bootstrap -hidden classes for that.

lucgeurts: Thanks, will make it mobile only with the hidden classes :) Also will do the same for the friesland version with dutch text.

PR #296, opened 2018-10-07
CA-376

lucgeurts:
Added Whatsapp sharing option in mobile view

### DIFF
--- a/CollAction/Frontend/app/global/ShareButtons.tsx
+++ b/CollAction/Frontend/app/global/ShareButtons.tsx
@@ -80,7 +80,7 @@ class ShareButtons extends React.Component<IShareButtonsProps, IShareButtonsStat
   }
 
   getWhatsappUrl() {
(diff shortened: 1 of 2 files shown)


CollAction/Frontend/app/global/ShareButtons.tsx
```diff
@@ -79,31 +79,42 @@ class ShareButtons extends React.Component<IShareButtonsProps, IShareButtonsStat
     return `https://www.linkedin.com/shareArticle?mini=true&url=${this.getUrl()}&title=${this.getTitle()}&source=${encodeURIComponent(window.location.origin)}`;
   }
 
+  getWhatsappUrl() {
+      return `whatsapp://send?text=${this.getTitle()} ${this.getUrl()}`;
```
logophobia: You have to escape spaces with %20 here (the space between ${this.getTitle()} and ${this.getUrl()}). Also, might be worth a try to see if it's possible to actually send something like <a href="${this.getUrl()}">${this.getTitle()}</a>.

PR #301, opened 2018-10-14
WIP: CA-495

dominicusplatus:


CollAction/Frontend/app/global/RichTextEditor.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react";
-import * as ReactQuill from "react-quill";
 import renderComponentIf from "./renderComponentIf";
+const ReactQuill = require("react-quill");
```
ljwboschker: What's wrong with the import statement? I'm not really familiar with this 'const' construct...

CollAction/Frontend/app/global/MobileNavDraw.tsx
```diff
@@ -12,23 +15,23 @@ interface IMobileNavDrawState {
...
+  } else {
     return <li><a href="/account/login">Login</a></li>;
   }
  }
 
 export default class MobileNavDraw extends React.Component<IMobileNavDrawProps, IMobileNavDrawState> {
 
-  constructor () {
-    super();
+  constructor (props?: IMobileNavDrawProps, context?: any) {
```
ljwboschker: I assume you had to do this because of the React upgrade. I did the same (but without the context parameter) when upgrading for the start project form. You might get merge conflicts when merging.

I think it's a good idea to merge master into this branch first to prevent the branches from falling apart any further :-)

CollAction/Frontend/app/global/MobileNavDraw.tsx
```diff
@@ -12,23 +15,23 @@ interface IMobileNavDrawState {
 function AccountManageNav(props) {
   if (props.isLoggedIn) {
     return <li><a href="/manage">Account</a></li>;
-  }else {
+  } else {
```
ljwboschker: Shall we use a newline in these "if else" constructs?

eg. 
```
if (true) {
  ...
}
else {
  ...
}
```

CollAction/Frontend/app/global/MobileNavDraw.tsx
```diff
@@ -45,7 +48,7 @@ export default class MobileNavDraw extends React.Component<IMobileNavDrawProps,
     return(
       <div id="draw-menu">
         <div id="draw-menu-close-button" onClick={() => self.close()}>
-          <i className="fa fa-times" aria-hidden="true"></i>
+          <FontAwesomeIcon icon={faTimes} />
```
ljwboschker: Maybe it's a good idea to use normal html here? So: `<i class="fa fa-times"></i>` instead of using a React component here? That way we keep it a little bit in line with the cshtml files.

CollAction/Views/Home/FAQ.cshtml
```diff
@@ -8,6 +8,7 @@
 
 @section styles {
     <link rel="stylesheet" href="~/css/faq.css" />
```
ljwboschker: This is an odd one here.... looks like legacy. Isn't the faq.scss already included by webpack?

CollAction/Views/Shared/_EmbeddedLayout.cshtml
```diff
@@ -16,10 +16,10 @@
         })(window, document, 'script', 'dataLayer', 'GTM-NTCKFKM');
     </script>
     <!-- End Google Tag Manager -->
-    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css"
+    <!-- <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css"
```
ljwboschker: Commented out code can be removed.

CollAction/Views/Shared/_Layout.cshtml
```diff
@@ -6,15 +6,16 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>@ViewData["Title"] - CollAction</title>
-    <script src="~/lib/google/tagmanager.js">
-    </script>
-    <environment names="Development">
+    <!-- <link href="~/font-awesome/css/font-awesome.css" rel="stylesheet" /> -->
```
ljwboschker: Comments can be removed.

CollAction/package.json
```diff
@@ -7,31 +7,51 @@
     "build": "webpack",
     "watch": "webpack -w"
   },
-  "author": "Stewart Matheson",
+  "author": "Collaction Team",
   "license": "ISC",
   "devDependencies": {
-    "@types/react": "^15.0.6",
-    "@types/react-dom": "^0.14.22",
+    "@fortawesome/fontawesome-free": "^5.3.1",
```
ljwboschker: Looks like we are using both CDN and include them in the javascript bundle... I think we should pick one and go with that. I prefer the CDN-solution.. can you check with @logophobia ? Not really sure why a include the fonts stuff in our own codebase, but there must have been a reason for that....

CollAction/webpack.config.js
```diff
@@ -1,32 +1,42 @@
 const webpack = require('webpack');
```
ljwboschker: The webpack config for webpack 4 can be lot smaller (it has a lot of sensible defaults). Check what's in master branch at the moment.

CollAction/package.json
dominicusplatus: font library has been removed, we now use v5 fontawesome which allows for using optimized version of fonts with inline svg generated for each import, it does not require entire font to be downloaded to client browser. It only requires javascript part to be bundled  then react components can use optimized version without downloading from CDN entire fonts of v4. It is small performance improvments for pages that are high traffic ( main page, footer etc. )

CollAction/Views/Shared/_EmbeddedLayout.cshtml
dominicusplatus: fixed in e2441e7

CollAction/Views/Shared/_Layout.cshtml
dominicusplatus: fixed in 81a0d8e

CollAction/Frontend/app/global/MobileNavDraw.tsx
dominicusplatus: It is due to performance improvement, v5 works different then v4, you can include single icon for page instead of downloading entire library.

CollAction/Frontend/app/global/MobileNavDraw.tsx
dominicusplatus: fixed in 81a0d8e

PR #308, opened 2018-11-17
Feature/ca 523

dominicusplatus:


dominicusplatus:
Sorry, mixed up with CA-527, it is correct number, will re-create request on correct branch

PR #310, opened 2018-11-25
CA-527 image size reduce

dominicusplatus:


CollAction/Services/Image/AmazonS3ImageService.cs
```diff
@@ -104,7 +107,12 @@ public async Task<Image<Rgba32>> UploadToImage(IFormFile fileUploaded)
                 using (MemoryStream ms = new MemoryStream())
                 {
                     await uploadStream.CopyToAsync(ms);
-                    return SixLabors.ImageSharp.Image.Load(ms.ToArray());
+                    var image = SixLabors.ImageSharp.Image.Load(ms.ToArray());
+                    var scaleRatio = GetScaleRatioForImage(image);
+                    image.Mutate(x => x
```
timstokman: Only mutate when scale factor !+ 1.0.

CollAction/Services/Image/AmazonS3ImageService.cs
```diff
@@ -118,6 +126,19 @@ private byte[] ConvertImageToPng(Image<Rgba32> image)
             }
         }
 
+        private double GetScaleRatioForImage(Image<Rgba32> image)
+        {
+            if(image.Width > _imageResizeThreshold || image.Height > _imageResizeThreshold){
```
timstokman: Formatting, use [allmann braces](https://en.wikipedia.org/wiki/Indentation_style#Allman_style)

timstokman: And spaces between ( and the "if"

CollAction/Services/Image/AmazonS3ImageService.cs
```diff
@@ -118,6 +129,22 @@ private byte[] ConvertImageToPng(Image<Rgba32> image)
             }
         }
 
+        private double GetScaleRatioForImage(Image<Rgba32> image)
+        {
+            if (image.Width > _imageResizeThreshold || image.Height > _imageResizeThreshold){
```
timstokman: The above line still has wrong braces formatting (must be on a newline)

CollAction/Services/Image/AmazonS3ImageService.cs
dominicusplatus: sory, fixed in 7a1eade

CollAction/Services/Image/AmazonS3ImageService.cs
timstokman: Maybe we start adding stylecop, then the compiler will start warning for badly formatted lines..

CollAction/Services/Image/AmazonS3ImageService.cs
dominicusplatus: sory, fixed in 7a1eade

CollAction/Services/Image/AmazonS3ImageService.cs
dominicusplatus: Sory, fixed in 7a1eade, will use Rider in future.

PR #314, opened 2018-11-29
CA-582 - performance, limit frontpage to 30 projects, add find more projects button

timstokman:


CollAction/Controllers/ProjectsController.cs
```diff
@@ -329,8 +329,7 @@ public async Task<JsonResult> FindProjects(int? categoryId, int? statusId)
                 default: statusExpression = (p => true); break;
             }
 
-            var projects = await _projectService.FindProjects(
-                Expression.Lambda<Func<Project, bool>>(Expression.AndAlso(projectExpression.Body, Expression.Invoke(statusExpression, projectExpression.Parameters[0])), projectExpression.Parameters[0]));
+            var projects = await _projectService.FindProjects(Expression.Lambda<Func<Project, bool>>(Expression.AndAlso(projectExpression.Body, Expression.Invoke(statusExpression, projectExpression.Parameters[0])), projectExpression.Parameters[0]), limit).ToListAsync();
```
ljwboschker: Not in scope for this PR, but using this Expression stuff looks really complicated...

CollAction/Services/Project/ProjectService.cs
```diff
@@ -83,38 +83,42 @@ public async Task<ProjectParticipant> GetParticipant(string userId, int projectI
             return await _context.ProjectParticipants.SingleOrDefaultAsync(p => p.ProjectId == projectId && p.UserId == userId);
         }
 
-        public async Task<IEnumerable<FindProjectsViewModel>> FindProjects(Expression<Func<Models.Project, bool>> filter)
+        public IQueryable<FindProjectsViewModel> FindProjects(Expression<Func<Models.Project, bool>> filter, int? limit)
         {
-            return await _context.Projects
+            IQueryable<Models.Project> projects = _context.Projects
```
jnaputi253: Is it possible for us to import the CollAction.Models namespace so that we can omit it from here?

timstokman: There's also a namespace "Collaction.Services.Project" which clashes with the model name. Which could rename one of them, but this solution is fine as well.

CollAction/Controllers/ProjectsController.cs
timstokman: I do have some ideas on how to refactor... Might put that in this PR.

CollAction/Services/Project/ProjectService.cs
jnaputi253: +1

PR #318, opened 2018-12-26
Updated start project section on homepage

lucgeurts:
Changed to the new design for project start on the homepage. Also cleaned out non-used code (e.g. carousel and related css code).


CollAction/Frontend/app/project/Carousel.tsx
```diff
@@ -1,212 +0,0 @@
-import * as React from "react";
```
timstokman: We don't need the carousel anymore at all? @ljwboschker Does that seem right?

ljwboschker: No, this was the last place it was used. Files can be removed; no use in keeping old stuff around.

CollAction/Frontend/app/project/styles/SimpleProjectCreate.scss
```diff
@@ -183,38 +183,4 @@
...
-    i {
-      float: right;
-    }
-
-    span {
-      margin-right: 10px;
-    }
-
-  }
-}
```
ljwboschker: I think this whole file can be removed; it was for the previous start account page.

lucgeurts: I've delete this file and tested it 👍

ljwboschker:
Approved.

PR #320, opened 2019-01-11
Add proper authorization header for Mailchimp

ljwboschker:
See https://developer.mailchimp.com/documentation/mailchimp/guides/get-started-with-mailchimp-api-3/#authentication. This seems to fix the problem.

Note: branche was created from prod15 tag. If this can go to production, we need to build a container from this branch.


ljwboschker:
Reviewed and approved by Tim (via Slack).

PR #325, opened 2019-02-17
Ca 580

timstokman:
Donation page


CollAction/Services/Donation/DonationService.cs
```diff
@@ -0,0 +1,165 @@
...
+
+                using (var request = new HttpRequestMessage()
+                {
+                    RequestUri = new Uri("https://api.stripe.com/v1/checkout/sessions"),
+                    Method = HttpMethod.Post,
+                    Content = new FormUrlEncodedContent(requestContent)
+                })
+                {
+                    request.Headers.Authorization = new AuthenticationHeaderValue("Basic", Convert.ToBase64String(Encoding.UTF8.GetBytes(_requestOptions.ApiKey + ":")));
+                    request.Headers.Add("Stripe-Version", "2019-03-14; checkout_sessions_beta=v1");
```
dominicusplatus: Does stripe require "date of API version" to process requests ? Could extract it to app--settings in future so that we can quickly change it following them, but I think it's not critical.

timstokman: We're using a beta stripe API here, because other flows don't comply with the upcoming SCA regulation (https://stripe.com/docs/strong-customer-authentication). That's why I'm doing this one in http, and the other one through the .net library. Otherwise we'd have to revise our entire page in half a year, now we only need to upgrade to the stable api version once it's ready.

CollAction/Frontend/app/global/styles/header.scss
```diff
@@ -1,171 +1,74 @@
...
 
-        li {
-            margin: 30px 20px;
-            font-size: 22px;
-
-            a {
-                color: $coll-action-green;
-            }
+        .navbar-toggle {
+            float: left;
```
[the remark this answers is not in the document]
timstokman: This is @ljwboschker 's part. Can you take a look at that @ljwboschker ?

CollAction/Frontend/app/donation/DonationBox.tsx
```diff
@@ -0,0 +1,290 @@
+import renderComponentIf from "../global/renderComponentIf";
+import * as React from "react";
+import IdealBox from "./IdealBox";
+import {StripeProvider, Elements, injectStripe} from 'react-stripe-elements';
+import * as Modal from 'react-modal';
+
+const InjectedIdealBox = injectStripe(IdealBox);
+
+const popupStyles = {
```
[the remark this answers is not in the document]
timstokman: Actually, with this react modal I can't. It's composed of multiple "divs", that all need to be styled separately, only way to do that is through a style object. You can try if you want to.

CollAction/Frontend/app/donation/DonationBox.tsx
```diff
@@ -0,0 +1,290 @@
...
+            console.log("Unable to redirect to checkout: " + responseBody);
+            this.setState({ showError: true, error: "Unable to redirect to checkout" });
+        }
+    }
+
+    onCloseDialog() {
+        this.setState({ showDialog: false });
+    }
+
+    setName(ev: React.ChangeEvent<HTMLInputElement>) {
```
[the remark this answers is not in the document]
timstokman: Done

CollAction/Frontend/app/donation/styles/donationbox.scss
```diff
@@ -0,0 +1,179 @@
...
+        color: $coll-action-green;
+    }
+
+    h1 {
+        color: white;
+    }
+}
+
+#donation-modal {
+    width: 26em;
```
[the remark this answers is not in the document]
timstokman: Done

CollAction/Frontend/app/global/styles/header.scss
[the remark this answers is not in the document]
ljwboschker: While I sympathise with the 'float is so 90s', the site still uses bootstrap 3 and the navbar styling there that I use, contains a 'float: right' in the navbar-toggle. A 'position: absolute' also works, but I am not sure if that is the right way to countermand the original bootstrap styling. What do you think @arnoutj?

CollAction/Frontend/app/donation/IdealBox.tsx
```diff
@@ -0,0 +1,71 @@
...
+        let sourceData = {
+            type: "ideal",
+            amount: this.props.amount * 100,
+            currency: "eur",
+            owner: {
+                name: this.props.userName,
+                email: this.props.userEmail
+            },
+            redirect: {
+                return_url: window.location.origin + "/donation/ThankYou"
```
ljwboschker: What happens if the user cancels the payment? Would be better to return to a dedicated URL that checks if there actually was a payment and then either redirect to the thank-you page or back to the donation page.

CollAction/Frontend/app/donation/IdealBox.tsx
```diff
@@ -0,0 +1,71 @@
...
+        };
+
+        let response = await this.props.stripe.createSource(sourceData);
+        if (response.error) {
+            console.log("Unable to start iDeal: " + response.error);
+            this.setState({ showError: true, error: "Unable to start iDeal" })
+            return;
+        }
+
+        let initializeUrl = `/donation/InitializeIdealCheckout?sourceId=${response.source.id}&name=${encodeURIComponent(this.props.userName)}&email=${encodeURIComponent(this.props.userEmail)}`;
```
ljwboschker: Just wondering about the Name field for iDeal payments. Isn't that unusual?

CollAction/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -118,6 +117,25 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                {
+                    b.Property<int>("Id")
+                        .ValueGeneratedOnAdd();
+
+                    b.Property<string>("EventData")
+                        .HasColumnType("json");
+
+                    b.Property<int>("Type");
+
+                    b.Property<string>("UserId");
```
ljwboschker: There is no maxlength here. Although it'll work fine, it might be wiser to add it here. Adding a MaxLenght or an index later will be troublesome on varchar(max) columns.

CollAction/Services/Donation/DonationService.cs
```diff
@@ -0,0 +1,165 @@
...
+
+            ApplicationUser user = await _userManager.FindByEmailAsync(email);
+            Customer customer = await GetOrCreateCustomer(name, email);
+
+            using (var client = new HttpClient())
+            {
+                var requestContent = new Dictionary<string, string>()
+                {
+                    { "success_url", $"{_siteOptions.PublicAddress}/Donation/ThankYou" },
+                    { "cancel_url", $"{_siteOptions.PublicAddress}/Donation/Donate" },
```
ljwboschker: That's interesting... it there no cancel_url for ideal payments? (see my earlier comment)

CollAction/Migrations/ApplicationDbContextModelSnapshot.cs
timstokman: It contains stripe json messages, no theoretical upper limit for that. I'm not sure what I'd set it to.

CollAction/Services/Donation/DonationService.cs
timstokman: This is a beta api, so a bit new.

CollAction/Frontend/app/donation/IdealBox.tsx
timstokman: I don't think there's a way you can cancel iDeal payments, other than clicking the back button or closing the page. My bank doesn't really offer one. If we have sources that could be cancelled, this would make sense. I initially had a more generic return url, but it only makes sense with more payment options.

timstokman: I'll add it back anyway, we can't test all banks, and it might change in the future.

timstokman: Did that, and fixed a bug in the payment flow (people could complete the iDeal payment, but we never "charged" the iDeal source).

PR #326, opened 2019-03-04
Ca 291 newsletter signup

arnoutj:
Replaced Mailchimp form in the footer with newsletter signup React component. 
This was necessary in order to show visual feedback after signup (success or failure).


CollAction/Frontend/app/global/NewsletterSignup.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  handleEmailChange(event) {
+    this.setState({
+      email: event.target.value,
+      subscriptionState: States.VOID
+    });
+  }
+
+  isValidEmail() {
+    return /[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,3}/.test(this.state.email);
```
timstokman: Validating the e-mail through a regex isn't a great idea, this will have a lot of false negatives. Maybe only if it contains the '@' symbol? https://davidcel.is/posts/stop-validating-email-addresses-with-regex/

CollAction/Frontend/app/global/styles/footer.scss
```diff
@@ -117,5 +89,76 @@
...
         }
     }
+
+
+    //Newsletter signup
+    .newsletter-signup {
+        position: relative;
+    }
+
+    .newsletter-signup__container {
```
timstokman: Bit of a weird naming convention. Why two underscores?

CollAction/Frontend/app/global/NewsletterSignup.tsx
```diff
@@ -0,0 +1,141 @@
...
+}
+
+interface NewsLetterSignupState {
+  subscriptionState: States;
+  email?: string;
+  signupStatusMessage?: string;
+}
+
+class NewsletterSignup extends React.Component<NewsLetterSignupProps, NewsLetterSignupState> {
+
```
timstokman: Inconsistent newline

CollAction/Frontend/app/global/NewsletterSignup.tsx
```diff
@@ -0,0 +1,141 @@
...
+    
+    fetch("http://collaction.us14.list-manage.com/subscribe/post", {
+      method: "POST",
+      body: formData,
+    })
+    .then(response => {
+      if (response.ok) {
+        this.showSignupSucces();
+      } else {
+        this.showSignupError();
```
timstokman: It might be the case that the http response actually has the signup error message, that might be useful to show as well (big difference between 'already subscribed', 'we suspect you might be a spammer', etc). Might be a good idea to check if that's the case.

timstokman:
Great work! Just had some minor comments.

You seem to have some build failures (see checks at the bottom of the pull request), might want to check that out. The latest master also has some buildfixes.

CollAction/Frontend/app/global/styles/footer.scss
arnoutj: It's the block-element-modifier (BEM) convention: https://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/. In this case newsletter-signup is the block and container the element. 

We used this at my previous employer for all websites. It may look weird in the beginning but works quite well. I woud like to introduce this convention in the whole frontend later on. But we can discuss that.

CollAction/Frontend/app/global/NewsletterSignup.tsx
arnoutj: Yes, thanks. I will update it.

CollAction/Frontend/app/global/NewsletterSignup.tsx
arnoutj: I'll fix it

CollAction/Frontend/app/global/NewsletterSignup.tsx
arnoutj: Fixed it.

arnoutj:
@timstokman the build failed on a problem that was on master as well, because it remained after I rebased onto master. I fixed it in my latest commit together with your other remarks.

CollAction/Frontend/app/global/NewsletterSignup.tsx
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    let formData = new FormData();
+    formData.append("MERGE0", this.state.email);
+    formData.append("u", this.props.mailchimpUser);
+    formData.append("id", this.props.mailChimpListId);
+
+    this.setState({ subscriptionState: States.LOADING });
+
+    fetch("http://collaction.us14.list-manage.com/subscribe/post", {
```
timstokman: This mailchimp account should probably be a configuration value.

CollAction/Views/Shared/_Footer.cshtml
```diff
@@ -59,35 +59,10 @@
...
-                <input type="submit" value="Subscribe" name="subscribe" id="mc-embedded-subscribe" class="button">
-              </div>
-            </div>
-          </form>
-        </div>
-        <!--End mc_embed_signup-->
+        <h5 class="header">@Localizer["Sign up to our newsletter"]</h5>
 
+        <!-- Mailchimp signup form -->
+        <div class="newsletter-signup" data-u="48e9b2f8f522cf59b9d5ffa8d" data-list-id="@NewsletterSubscriptionServiceOptions.Value.MailChimpNewsletterListId"></div>
```
timstokman: 48e9b2f8f522cf59b9d5ffa8d probably needs to be fed from a configuration value, just like @NewsletterSubscriptionServiceOptions.Value.MailChimpNewsletterListId

arnoutj: I added it as a new variable @NewsletterSubscriptionServiceOptions.Value.MailChimpUserId

CollAction/Frontend/app/global/NewsletterSignup.tsx
arnoutj: Done

CollAction/Frontend/app/global/NewsletterSignup.tsx
arnoutj: Updated it to a check for just '@' symbol as you suggested @timstokman

CollAction/Services/Newsletter/NewsletterSubscriptionService.cs
```diff
@@ -59,6 +60,7 @@ public NewsletterSubscriptionService(IOptions<NewsletterSubscriptionServiceOptio
             _newsletterListId = options.Value.MailChimpNewsletterListId;
             _jobClient = jobClient;
             _logger = logger;
+            _userId = options.Value.MailChimpUserId;
```
timstokman: If it's not used in this service, it should be added as a variable.

CollAction/Frontend/app/global/NewsletterSignup.tsx
timstokman: Not pushed yet?

CollAction/Services/Newsletter/NewsletterSubscriptionService.cs
arnoutj: Do you mean as a variable in Index.cshtml?

timstokman: No, I mean _userId. It looks like a new variable, and it's not used anywhere, so it's redundant.

CollAction/Frontend/app/global/NewsletterSignup.tsx
arnoutj: I don't see the problem. New line before and after class definition like in other components..

CollAction/Services/Newsletter/NewsletterSubscriptionService.cs
arnoutj: I see. I removed it

CollAction/Frontend/app/global/NewsletterSignup.tsx
timstokman: Here's a typical react example: https://reactjs.org/docs/components-and-props.html

This is just making sure the formatting is consistent with typical typescript/react code.

timstokman: Would be good to add a newline between the interface definition and the class as well.

arnoutj: Ok I updated it.

ljwboschker:
I think be can remove this PR, right?

arnoutj:
Yes!

PR #332, opened 2019-03-23
CA-640 - user event tracking endpoint

timstokman:


timstokman:
The events are posted through tag manager yes.

PR #336, opened 2019-03-31
CA-591

dominicusplatus:
Remove image dimension limits from admin console


dominicusplatus:
Sorry, bad branch/code, removed changes which were for CA-559, this changes now are valid for CA-591 CollAction.org: remove "friesland" filter on find project page

CollAction/Migrations/20190407192822_CA_591_remove_category_friesland.cs
```diff
@@ -0,0 +1,17 @@
...
+    public partial class CA_591_remove_category_friesland : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("DELETE FROM \"Categories\" WHERE \"Name\" = 'Friesland'");
+        }
+
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("DELETE FROM \"Categories\" WHERE \"Name\" = 'Friesland'");
```
timstokman: On "Down" it needs to recreate the category. The "Down" action is supposed to reverse the migration, so we can always roll back database actions.

dominicusplatus: I understand that, but do you want to create an update/insert with that data ? I have been following similiar issue with update that was refered, which does the same for Up/Down

dominicusplatus: For some reason I do not have that entry in my local DB from master, so I do not know how to set the color column for it 🤦‍♂️ as it is stored as an integer.

timstokman: You can check the Friesland branch to see which data you'd need to re-insert. The list of categories and the exact data can be found in the function "CreateCategories" in ApplicationDbContext.cs.

Although, in this case, it's probably good enough to keep it empty. Doubt we'd need to roll this.

dominicusplatus: Added Down insert with default color in ebbee14

PR #344, opened 2019-05-16
Ca 415 670

timstokman:
* Add "templated" e-mails, to make our transactional e-mails more managable (CA-415)
* Add a thank-you e-mail for donations (CA-670)

Because CA-618 hasn't been merged yet, it includes that change since it's necessary.

There're still a few pending questions, so this isn't ready to be reviewed yet.


timstokman:
Branch name is in the wrong format, closing, and opening a new one

PR #345, opened 2019-05-17
Ca 321 lazy signup

ljwboschker:
- Front-end now shows an e-mail input field if you want to participate while not logged in
- Participation logic is moved from ProjectsService to it's own service
- E-mail logic is moved into separate helper class (there are 6 scenarios resulting in 5 different mails)
- Use password reset token mechanism to finish account registration after lazy sign-up


timstokman:
If you want, I can merge the ViewRenderer for you from CA-415. That might make a lot of the e-mail generation nicer, and will make the eventual merge easier.

CollAction/Controllers/AccountController.cs
```diff
@@ -328,6 +328,54 @@ public IActionResult ResetPasswordConfirmation()
             return View();
         }
 
+        [HttpGet]
+        [AllowAnonymous]
+        public IActionResult FinishRegistration(string email, string code)
+        {
+            var model = new FinishRegistrationViewModel { Email = email, Code = code };
```
timstokman: If you put the model directly in the parameter list (instead of string email, string code), you can skip this line

CollAction/Services/Project/IParticipantsService.cs
```diff
@@ -0,0 +1,25 @@
...
+    public interface IParticipantsService
+    {
+        Task<AddParticipantResult> AddAnonymousParticipant(int projectId, string email);
+        Task<AddParticipantResult> AddLoggedInParticipant(int projectId, string userId);
+        Task RefreshParticipantCountMaterializedView();
+        Task<ProjectParticipant> GetParticipant(string userId, int projectId);
+        Task<string> GenerateParticipantsDataExport(int projectId);
+    }
+
+    public enum AddParticipantScenario
```
timstokman: I think the guideline is that separate interfaces/enums/classes should be in separate files

CollAction/Services/Project/ParticipantsService.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using CollAction.Data;
+using CollAction.Models;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.EntityFrameworkCore;
+
+namespace CollAction.Services.Project
+{
+  public class ParticipantsService : IParticipantsService
```
timstokman: Wrongly indented here

CollAction/Services/Project/ParticipantsService.cs
```diff
@@ -0,0 +1,142 @@
...
+
+            var result = new AddParticipantResult();
+            var user = await _userManager.FindByEmailAsync(email);
+            if (user == null)
+            {
+                var creationResult = await _userManager.CreateAsync(new ApplicationUser(email));
+                if (!creationResult.Succeeded)
+                {
+                    var errors = string.Join(',', creationResult.Errors.Select(e => $"{e.Code}: {e.Description}"));
+                    throw new InvalidOperationException($"Could not create new unregisterd user {email}: {errors}");
```
timstokman: unregisterd is not a word

CollAction/Views/Projects/Details.cshtml
```diff
@@ -11,6 +12,10 @@
     var canTrack = consentFeature.CanTrack;
 
     ViewData["Title"] = @Model.Project.Name + " Details";
+
+    var user = await userManager.GetUserAsync(User);
```
timstokman: This looks like code that should be in the viewmodel, controller or service

CollAction/Controllers/AccountController.cs
ljwboschker: Unfortunately, it will conflict with the POST method that has the same name.

CollAction/Services/Project/IParticipantsService.cs
ljwboschker: Done

CollAction/Services/Project/ParticipantsService.cs
ljwboschker: Done

CollAction/Services/Project/ParticipantsService.cs
ljwboschker: Done

CollAction/Views/Projects/Details.cshtml
ljwboschker: I moved this to the controller, and return it to the view using ViewData since it's not really a part of the project details.

ljwboschker:
@timstokman Since the slow fashion project starts next week, I think it's best to add the e-mail templates later. I don't have a lot of time for the rest of this weekend...

timstokman:
@ljwboschker I'll add that to my pull request then. Mine is a little less urgent.

PR #346, opened 2019-05-17
CA-670 & CA-415

timstokman:
Adds a thank-you e-mail for donations (normal and recurring ones), and adds a templating system for e-mails.


CollAction/Views/Emails/DonationThankYou.cshtml
```diff
@@ -0,0 +1,17 @@
+﻿@model CollAction.Models.EmailViewModels.DonationThankYouEmailViewModel
+Hi @Model.Name!,<br>
+<br>
```
ljwboschker: Why no paragraphs here? Eg. <p>

timstokman: Done.

Mostly needed the paragraph tags because you can't use raw text inside if-statements. They're better too semantically, so I changed the donation e-mail as well.

PR #347, opened 2019-05-20
CA-673

timstokman:
Store referer, utm in session storage


CollAction/Frontend/app/global/RegisterReferer.ts
```diff
@@ -0,0 +1,29 @@
+﻿function parseUtmParameters() : string {
+    const utmKeys = ["utm_source", "utm_medium", "utm_campaign", "utm_term", "utm_content"];
+    const urlParams = new URLSearchParams(window.location.search);
```
ljwboschker: Whatch out; URLSearchParams isn't supported in IE11.

ljwboschker: You can either use a polyfill or your own function like https://gomakethings.com/getting-all-query-string-values-from-a-url-with-vanilla-js/

timstokman: I added a polyfill

PR #354, opened 2019-05-31
Add disqus component to project details

ljwboschker:
WIP: this PR is to enable deployment of a version of CollAction with Disqus, to see if this is usable for CollAction.


CollAction/Views/Projects/Details.cshtml
```diff
@@ -209,7 +212,51 @@
                 </div>
             </div>
         }
+    </div>
+    
+    @if (canTrack)
+    {
+        <div class="container">
```
arnoutj: What is the use of this empty container?

CollAction/Views/Projects/Details.cshtml
```diff
@@ -209,7 +212,51 @@
...
+                    <div class="col-xs-12">
+                        <div id="disqus_thread"></div>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        var identifier = $"{@Model.Project.Name}/{@Model.Project.Id}";
+        var script = $"https://{@SiteOptions.Value.DisqusSite}.disqus.com/embed.js";
+        <script>
```
arnoutj: I think it is better to put script tags at the bottom of the page, since JS loading blocks the DOM rendering of the rest of the page right?

CollAction/Views/Projects/Details.cshtml
ljwboschker: I added the <hr> in it to disqus_thread div column.

CollAction/Views/Projects/Details.cshtml
ljwboschker: I already moved it to disqus.ts and include it in the general frontend bundle.

PR #359, opened 2019-06-14
CA-675 - client-side e-mail validation lazy sign-up

timstokman:


CollAction/Frontend/app/project/styles/ProjectDetails.scss
```diff
@@ -244,4 +244,8 @@
     flex-direction: column;
     align-items: center;
     text-align: center;
+
+    .error {
+        color: $coll-action-green;
```
ljwboschker: Other forms show a red error message. Shouldn’t we keep it the same?

timstokman: The donation error colors were green as well. But you're right, red is probably better, I did ensure that all our "error" classes are now red though.

PR #365, opened 2019-06-29
CA-682 - fix caching front-end

timstokman:
* I added caching headers to our static assets that cache them for a week
* I added the hash of frequently changing assets to the query string, so the browser sees them as different things (only the javascript, not the images)
* Moved some things to the bottom to speed up page loading


CollAction/Startup.cs
```diff
@@ -285,7 +287,14 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
 
             app.UseAuthentication();
 
-            app.UseStaticFiles();
+            app.UseStaticFiles(new StaticFileOptions()
+            {
+                OnPrepareResponse = ctx =>
```
ljwboschker: This will cache all static resources, including images, fonts and style sheets. But you only hash the JavaScript.

CollAction/Services/HashAssetService/HashAssetService.cs
```diff
@@ -0,0 +1,46 @@
...
+        public string HashAssetBase64(string[] location)
+            => Base64UrlEncoder.Encode(HashAsset(location));
+
+        private byte[] Hash(string location)
+        {
+            using (var sha256 = SHA256.Create())
+            {
+                using (var stream = File.OpenRead(location))
+                {
+                    return sha256.ComputeHash(stream);
```
ljwboschker: npm run watch won’t work anymore without hard cash resets (or backend restart) this way.

CollAction/Startup.cs
timstokman: The stylesheets are, as far as I can see, packed into the javascript. So they're included.

The images are changed for a few reasons:
1 Resizing to optimize performance
2 Actually changing the image

For 1, I would suggest that it's fine to keep caching the old version for a week. 2 is rare, and only happens on redesign, I would suggest we do a rename if that happens.

timstokman:
Looks like webpack has a similar feature like you said luuk (https://stackoverflow.com/questions/35176489/what-is-the-purpose-of-webpack-hash-and-chunkhash/52786672#52786672). But it looks like it might be harder to integrate it to the _Layout than the HashAssetService solution. But feel free to take a look.

PR #369, opened 2019-07-12
Ca 696 update register page

ljwboschker:
Intercepts the DuplicateUserName error when registering a new user with Identity framework.

Because the story asks for HTML mark-up in the message to the user, I filter out the DuplicateUserName error from the error list returned when calling the UserManager, and use specific check of the ViewData.ModelState in the razor template.

The reactivation e-mail can contain two links:
* FinishRegistration if the account is not complete yet, most likely because it was registered as an anonymous participant when committing to a project
* PasswordReset link if the account is complete

Also updated the text in the password reset e-mail. It wasn't really proper English...


CollAction/Views/Account/Register.cshtml
```diff
@@ -19,7 +19,19 @@
         <div class="row">
             <div class="col-xs-12 col-md-6 col-md-offset-3">
                 <form asp-controller="Account" asp-action="Register" asp-route-returnurl="@ViewData["ReturnUrl"]" method="post">
+                    <!-- <div asp-validation-summary="All" class="text-danger"></div> -->
```
timstokman: No commented out code please

timstokman:
What's up with the Gatsby cms file?

CollAction/Views/Emails/ResetPassword.cshtml
```diff
@@ -1,10 +1,17 @@
...
+<p>
 Warm regards,<br>
-The CollAction Team<br>
-<br>
-PS. If you've haven't requested this e-mail, you can ignore it.
\ No newline at end of file
+The CollAction Team
+</p>
+<p>
+PS. If you did not request a password reset, you can ignore this mail. However, it might be a good idea to <a href="https://howsecureismypassword.net/">check your password's strenght</a>!
```
timstokman: "strenght" should be "strength"

CollAction/Views/Account/Register.cshtml
ljwboschker: Removed

CollAction/Views/Emails/ResetPassword.cshtml
ljwboschker: Good catch!

ljwboschker:
Gatsby stuff was accidentally committed. It's gone now.

CollAction/Views/Emails/ResetPassword.cshtml
```diff
@@ -1,10 +1,17 @@
...
+<p>
 Warm regards,<br>
-The CollAction Team<br>
-<br>
-PS. If you've haven't requested this e-mail, you can ignore it.
\ No newline at end of file
+The CollAction Team
+</p>
+<p>
+PS. If you did not request a password reset, you can ignore this mail. However, it might be a good idea to <a href="https://howsecureismypassword.net/">check your password's strength</a>!
```
timstokman: Maybe we should add a password-strength-indicator on our registration pages instead of this line. We could add a follow-up ticket for it.

CollAction/Controllers/AccountController.cs
```diff
@@ -253,18 +285,19 @@ public async Task<IActionResult> ForgotPassword(ForgotPasswordViewModel model)
                 var user = await _userManager.FindByEmailAsync(model.Email);
                 if (user == null)
                 {
-                    // Don't reveal that the user does not exist
```
timstokman: I'd prefer to keep this comment, stops people from adding validation

ljwboschker: Re-added

PR #403, opened 2020-05-31
CA-842 - second category project creation

timstokman:


timstokman:
@ljwboschker This might be a good simple solution until we have time to add that component. The crowdaction start form isn't used that often, so I'd say this would be a decent temporary solution.

PR #406, opened 2020-06-05
CA-833 - fix datepicker safari

JBijker:


Frontend/src/pages/crowdactions/Create/form.ts
```diff
@@ -61,6 +62,19 @@ var minimumDate = new Date();
 var maximumDate = new Date();
 maximumDate.setMonth(minimumDate.getMonth() + 12);
 
+//Doesn't seem to work...
```
timstokman, line 65: Remove the commented out code. No commented out code in the final version (that's what we have git for).

JBijker: Check!

(On a side note, I think it makes sense to leave the comments in there, see ticket on Jira. Form validation would still be good to add, and the comments should do that, it's just not working)

timstokman (approved):

PR #407, opened 2020-06-12
Ca 851

timstokman:


CollAction/Controllers/GraphQlController.cs
```diff
@@ -140,7 +140,7 @@ private async Task<ExecutionResult> Execute(string query, string? operationName,
                 UserContext = new UserContext(User, context, serviceProvider),
                 ComplexityConfiguration = new ComplexityConfiguration()
                 {
-                    MaxDepth = 20
+                    MaxDepth = 21
```
ljwboschker, line 143: Magic number? Why is it 21 now? What does this mean? Use Constant With meaningful name

CollAction/GraphQl/Queries/CrowdactionGraph.cs
```diff
@@ -98,6 +104,23 @@ public CrowdactionGraph(IEfGraphQLService<ApplicationDbContext> entityFrameworkG
...
+                    if (instagramUser != null)
+                    {
+                        return c.GetUserContext()
+                                .ServiceProvider
+                                .GetRequiredService<IInstagramService>()
+                                .GetItems(instagramUser, c.CancellationToken);
+                    }
+                    else
+                    {
+                        return Task.FromResult(Enumerable.Empty<InstagramWallItem>());
```
ljwboschker, line 121: Why not use async / await delegate here?

Frontend/src/pages/crowdactions/Create/form.ts
```diff
@@ -84,6 +86,7 @@ export const validations = Yup.object({
   hashtags: Yup.string()
     .max(30, 'Please keep the number of hashtags civil, no more then 30 characters')
     .matches(/^[a-zA-Z_0-9]+(;[a-zA-Z_0-9]+)*$/, 'Don\'t use spaces or #, must contain a letter, can contain digits and underscores. Separate multiple tags with a colon \';\''),
+  instagramUser: Yup.string(),
```
ljwboschker, line 89: Min / Max length check like in back end is missing

CollAction/GraphQl/Queries/CrowdactionGraph.cs
timstokman: Because it adds overhead. Just straight-up passing the task object down is a lot more efficient.

Frontend/src/pages/crowdactions/Create/form.ts
timstokman: A minimum check is not necessary. If the string is empty, it will send a null, which is also valid on the backend. I added the max check though.

CollAction/Controllers/GraphQlController.cs
timstokman: Should be fixed

PR #415, opened 2020-06-27
CA 776 - fix e-mails on project completion / automate project success failure

timstokman:
There were multiple issues:
* The job at the end didn't include all the information necessary in the query
* The job setup code was checking for the wrong conditions
* View couldn't be rendered w/o HttpContext, which isn't available in a hangfire job. So I introduced RazorLight.

Included in this is:
* Optimization of comment seeding
* Added integration tests for e-mail rendering
* Remove ConfigureAwait from the unit test project (bit useless)


CollAction/CollAction.csproj
```diff
@@ -4,7 +4,33 @@
     <UserSecretsId>aspnet-CollAction-29a5d7b8-3530-4482-9ed6-572bd6c178fa</UserSecretsId>
     <LangVersion>8.0</LangVersion>
     <Nullable>enable</Nullable>
+    <PreserveCompilationReferences>true</PreserveCompilationReferences>
+    <PreserveCompilationContext>true</PreserveCompilationContext>
   </PropertyGroup>
+  <ItemGroup>
```
ljwboschker, line 10: Why specifically remove them, and add them below here? Probably something VS did?

.gitignore
```diff
@@ -1,6 +1,7 @@
 .vs
 *.user
 *.cache
+*.received.txt
```
ljwboschker, line 4: This is in the git-ignore, yet there are a lot of .received.txt files in this comment?

timstokman: If you look, I only added .verified.txt. The test works as follows:
* We specify the document we expect (verified.txt)
* The unit test generates *.received.txt
* They must be equal

Verify.Xunit is a xunit extension specific for verifying documents easier (instead of using gigantic inline strings).

CollAction/CollAction.csproj
timstokman: This is what happens if you add them as embeddedresource (which is what razorlight uses). Pretty sure what visual studio did is correct, bit weird, but if you remove parts it doesn't work.

.gitignore
timstokman: The integration tests are basically a way to test if the e-mail templates are generated you'd expect them to. Might be a bit overkill, but we can't really test IEmailSender directly (because you'd need to e-mail someone, and it'd cost us money to run the tests).

PR #416, opened 2020-06-28
CA 844 - Anonymous comments

timstokman:
This will allow people to leave anonymous comments with only a name on projects. They need to be approved before they appear. In the admin UI there'll be a place where you can approve anonymous comments. Comments for logged-in users are auto-approved, but can be removed afterwards.


timstokman:
@ljwboschker Added a description. You can always look at the ticket in JIRA at what it says.

PR #418, opened 2020-06-28
Ca 836 rich text errors

ljwboschker:
Upgrade the underlying component
Increase placeholder size so user can now click on the whole editor to activate it (instead of on the placeholder text)
Add styling hacks to show the red line at the bottom


Frontend/src/components/RichTextEditorFormContol/RichTextEditorFormControl.tsx
```diff
@@ -46,7 +46,11 @@ export class RichTextEditorFormControl extends React.Component<
             height: this.props.height || "calc(var(--spacing) * 20)",
             overflow: "scroll",
             borderBottom: "1px solid var(--c-grey-d20)"
-          }
+          },
+          placeHolder: {
+            // make the whole editor clickable
+            height: "calc(100% - 55px)"
```
arnoutj, line 52: What is the reason for this 55px?

ljwboschker: It's determined inside the mui-rte component; I couldn't find the exact rule... so I decided to put the exact value in here.

PR #420, opened 2020-07-03
Ca 859 test select

ljwboschker:
- min health host count allows the new container to come online before the old one finishes draining, allowing for faster switching
- netlify preview branches won't work due to cors limitation. Add configurable list of partial hosts names that allow --collaction.netlify.app


CollAction/Startup.cs
```diff
@@ -34,6 +34,7 @@
 using Serilog;
 using Stripe;
 using System;
+using System.Linq;
```
timstokman, line 37: Undo this?

PR #421, opened 2020-07-04
CA-861 - Fix getCookie so we can get client_id correctly

timstokman:
The ga client_id is needed for the analytics

### DIFF
--- a/Frontend/src/providers/AnalyticsProvider.tsx
+++ b/Frontend/src/providers/AnalyticsProvider.tsx
@@ -23,8 +23,16 @@ const SEND_EVENT = gql`
 
 // Get cookie by name
 const getCookie = (name: string) => {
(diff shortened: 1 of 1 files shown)


Frontend/src/providers/AnalyticsProvider.tsx
```diff
@@ -23,8 +23,16 @@ const SEND_EVENT = gql`
...
 // Get cookie by name
 const getCookie = (name: string) => {
-    var v = document.cookie.match(`(^|;) ?'${name}'=([^;]*)(;|$)`);
-    return v ? v[2] : null;
+  const found =
+    document.cookie
+            .split('; ')
+            .map(row => row.split('='))
+            .find(row => row[0] === name);
+  if (found) {
```
arnoutj, line 31: You could also do this:`return found ? found[1] |: null`

PR #423, opened 2020-07-10
CA-863 - search users in admin UI

timstokman:


Frontend/src/components/Admin/Users/AdminListUsers.tsx
```diff
@@ -14,14 +14,16 @@ export default () => {
     const [toDelete, setToDelete] = useState<IUser | null>(null);
     const [info, setInfo] = useState<string | null>(null);
     const [error, setError] = useState<string | null>(null);
+    const [search, setSearch] = useState("");
```
arnoutj, line 17: Type is missing

Frontend/src/components/Admin/Users/AdminListUsers.tsx
```diff
@@ -75,6 +77,7 @@ export default () => {
                 <Button onClick={() => setDeleteDialogOpen(false)}>Cancel</Button>
             </DialogActions>
         </Dialog>
+        <TextField label="Search user" type="text" value={search} onChange={(ev) => { setSearch(ev.target.value); setPage(0); }} />
```
arnoutj, line 80: I would move these to one combined function to keep the jsx clean

Frontend/src/components/Admin/Users/AdminListUsers.tsx
timstokman: The type is inferred in this case, it can only be a string.

In the cases above, I had to force it to be a certain type.

Frontend/src/components/Admin/Users/AdminListUsers.tsx
timstokman: Alright, refactored a few other similar things into functions as well then.

PR #425, opened 2020-07-24
Add social media sharing buttons

ljwboschker:
Use react-share lib for social media sharing.


Frontend/src/pages/crowdactions/Detail/SocialMediaSharing.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { EmailIcon, EmailShareButton, FacebookIcon, FacebookShareButton, FacebookShareCount, LinkedinIcon, LinkedinShareButton, TumblrIcon, TumblrShareButton, TumblrShareCount, TwitterIcon, TwitterShareButton, WhatsappIcon, WhatsappShareButton } from 'react-share';
+import { ICrowdaction } from '../../../api/types';
+
+import styles from './SocialMediaSharing.module.scss';
+
+const SocialMediaSharing = ({ crowdaction }: any) => {
+
```
timstokman, line 8: Unnecessary empty line.

Frontend/src/pages/crowdactions/Detail/SocialMediaSharing.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { EmailIcon, EmailShareButton, FacebookIcon, FacebookShareButton, FacebookShareCount, LinkedinIcon, LinkedinShareButton, TumblrIcon, TumblrShareButton, TumblrShareCount, TwitterIcon, TwitterShareButton, WhatsappIcon, WhatsappShareButton } from 'react-share';
+import { ICrowdaction } from '../../../api/types';
+
+import styles from './SocialMediaSharing.module.scss';
+
+const SocialMediaSharing = ({ crowdaction }: any) => {
```
timstokman, line 7: Instead of casting, you could add:

```
interface ISocialMediaSharingProps {
    crowdaction: ICrowdaction;
}
```

And then:

`const SocialMediaSharing = ({ crowdaction }: ISocialMediaSharingProps) => {`

For an example, see: https://github.com/CollActionteam/CollAction/blob/a454ebfaded866cf8e77cfc3728752a103638c0c/Frontend/src/components/Profile/CrowdactionParticipating.tsx

Then we're actually checking types instead of casting, less chance of runtime issues.

Frontend/src/pages/crowdactions/Detail/SocialMediaSharing.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
timstokman, line 1: Might be good to place this in the components directory, it's generally reusable

Frontend/src/pages/crowdactions/Detail/SocialMediaSharing.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import styles from './SocialMediaSharing.module.scss';
+
+const SocialMediaSharing = ({ crowdaction }: any) => {
+
+  const action = crowdaction as ICrowdaction;  
+  const shareUrl = window.location.href;
+
+  return (
+    <div className={styles.shareButtons}>
```
timstokman, line 13: These components call the facebook/... API, there are likely privacy implications. A number of things are needed:

* Use the useConsent hook to check for social permissions before adding these components
* Check the privacy policy to check if we're allowed to do this (probably yes, but good to check)
* Might be good to ask ourselves if we really want the counters, might be easier to disable them. Something to discuss at the call?

Frontend/src/pages/crowdactions/Detail/SocialMediaSharing.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <LinkedinIcon size={32} round />
+      </LinkedinShareButton>
+
+      <TumblrShareButton url={shareUrl} title={action.name} caption={action.proposal} tags={action.tags.map(tag => tag.tag.name)}>
+        <TumblrIcon size={32} round />        
+      </TumblrShareButton>
+      <TumblrShareCount url={shareUrl}>
+        {count => count}
+      </TumblrShareCount>
+
```
timstokman, line 35: I would've expected instagram here somewhere, giving how much it's used to promote collaction

Frontend/src/pages/crowdactions/Detail/SocialMediaSharing.tsx
ljwboschker: Counters are disabled, since they are not in the design either. Components are not calling any outside APIs.

Frontend/src/pages/crowdactions/Detail/SocialMediaSharing.tsx
ljwboschker: There is no Instagram share button (or API) for websites. At least I couldn't find one.

Frontend/src/pages/crowdactions/Detail/SocialMediaSharing.tsx
ljwboschker: Moved to higher level, and added to commit page as well. Noticed there already was a component. I replaced it with this one (has more options).

PR #426, opened 2020-12-12
CA-868 fix front-end validation

ljwboschker:
Simplified date entry on project creation form. Dates must now always be entered (with plain text input) as d/M/yyyy. At least it will work in every browser, no matter what localisation settings.

This also makes entry easier then de UI date-selector, and I readded the client-side validation so user get validation feedback on the dates before project is submitted.

### DIFF
--- a/Frontend/package.json
+++ b/Frontend/package.json
@@ -3,7 +3,6 @@
   "private": true,
   "dependencies": {
     "@apollo/client": "^3.0.0-beta.50",
-    "@date-io/moment": "1.x",
     "@fortawesome/fontawesome-svg-core": "^1.2",
     "@fortawesome/free-brands-svg-icons": "^5.12",
     "@fortawesome/free-regular-svg-icons": "^5.12",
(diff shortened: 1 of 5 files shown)


timstokman:
Managed to get a little closer to the bottom of this issue. I suspect it's actually related to validation issues:
* If you leave the target date to the default, it'll trigger a silent validation failure (with the dates being sent as empty). Here the problem is with how validation is setup.
* If you fill in the tags wrong, it'll also trigger a silent validation error

ljwboschker:
Proposed solution fixes date validation.

timstokman (approved):

Frontend/package.json
```diff
@@ -61,7 +60,7 @@
     "typescript": "^3.8"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "export REACT_APP_BACKEND_URL=https://localhost:44301 && react-scripts start",
```
timstokman, line 63: This would make things a little difficult to run on windows. Best run the start command as:

REACT_APP_BACKEND_URL=https://localhost:44301 npm run start

instead.

ljwboschker: Removed the env export

